Issue #2, opened 2026-08-24
面試途中 AI 失憶並重新開場


nosba0957:
我使用 45 分鐘的面試，題目選擇 Min Stack，大約進行 10 分鐘 (剩餘時間 35 分鐘) 時，我講話到一半 AI 就直接從開頭重來並且打招呼，並也忘記我剛才說了什麼。
註：測試時在穩定的網路中，平時使用是不會斷線的。

alanhc:
The cause is the Live API's connection lifetime rather than anything about the network, which is consistent with the stable connection you describe. Google closes a Live WebSocket after about ten minutes however much of the session is left to run, and that matches the ten minute mark here.

What turns that into amnesia rather than a pause is what happens next. `run_room` treated the hang-up as the end of the interview: it closed the room and let the supervisor restart. The restart sends `boot.greeting` unconditionally, against a Gemini session that has no history, so the interviewer introduces itself again and has lost everything said before. `src/agent/prompts.rs` does forbid a second greeting, but it says "in this conversation", and after the reconnect that is not the same conversation.

The setup message has asked for `sessionResumption` since it was written, which is why this looks like it was already handled. That half is only a request for handles: nothing read the `sessionResumptionUpdate` frames carrying them, and `open_live_session` had nowhere to put one, so the feature was present at the call site and did nothing on the wire.

#3 captures the handle and reconnects with it, and does not re-send the greeting.

One caveat worth stating plainly: I have not yet watched a real ten minute close and resume end to end. The wire format and the reader are covered by tests, and `check-gemini` still succeeds against the live API, but the reconnect itself has only been reasoned about. A run past ten minutes should log `Gemini session resumed (1 so far)` and greet exactly once. @nosba0957, if you are able to repeat your Min Stack session against that branch, that would be the confirmation this needs.

PR #3, opened 2026-08-25
Resume the interview when Gemini closes the socket

alanhc:
Gemini closes a Live connection after about ten minutes whatever the session has left to run, and the interview ended with it: `run_room` logged the hang-up, closed the room, and let the supervisor restart. The restart sends `boot.greeting` unconditionally against a session with no history, so a candidate twenty minutes into a forty-five minute interview was greeted a second time by an interviewer that had forgotten the first ten. The prompt already forbids a second greeting, but it says "in this conversation", and the new session is not that conversation.

The setup message had asked for `sessionResumption` since it was written. That half is only a request for handles; nothing read the `sessionResumptionUpdate` frames carrying them, and nothing could have spent one, because `open_live_session` had nowhere to put it. The feature looked present at the call site and did nothing on the wire.

`parse_server_message` now returns the handle and the `goAway` warning beside the events, and the reader task keeps the handle in an `Arc<Mutex<..>>` the session exposes. It is held there rather than sent through the event channel because of when it is wanted: after the socket closed and the last events drained, when the channel is finished and the caller has to decide whether it can resume at all.

A handle is taken only from an update the server marked `resumable`. Mid-turn updates carry one that is refused on reconnect, so accepting those would replace a working checkpoint with a broken one. A resumed connection also seeds itself with the handle it resumed from, since the server does not always re-issue one straight away and losing it would turn the next close into the cold start this exists to prevent.

The reconnect does not re-send the greeting, and closes the turn ids first, for the reason an interruption closes them: left open, the next thing either party says appends to an utterance that was cut off, and the panel and the report both read the two as one.

## What the first push got wrong

Reading the close is not enough on its own. Every arm of the loop that writes to Gemini propagated its error, and a socket the server has closed fails the next send long before `next_event` drains and reports it. Audio flushes every hundred milliseconds, so in practice the write side always won: the interview ended from a `?` on a dropped frame, and the arm that resumes it never ran. The first version of this PR would not have fixed the reported bug. Those four writes now log and carry on, leaving the close to the one arm that can explain it. A tenth of a second of audio is the cost.

`connect_async` had no timeout either, where the exchange over it had `SETUP_TIMEOUT`. A connect that never answers hangs `run_room`'s loop, which is also how the interview notices its candidate leaving and how its own deadline fires, so a stalled reconnect could strand a room that should have ended. `CONNECT_TIMEOUT` bounds the socket the way `SETUP_TIMEOUT` bounds what crosses it.

Both of these came from @cubic-dev-ai's review on the first push. Thanks.

## Testing

`take_resume_attempt` holds the part of resuming that can be judged: `cargo test` cannot open a Gemini session, but the ceiling and the counter are arithmetic. Three tests name both sides of the boundary, that a refusal spends nothing, and that a resume spends exactly one — counting from three, so a multiply could not pass for an add. The reconnect itself stays inline in `run_room`, which `.cargo/mutants.toml` already excludes, rather than behind a name that would add a mutant no test can kill.

The mutation gate is why: the first push missed four mutants, all of them in the function that held this logic, which is a fair report of it having had no test at all.

## Verification

`cargo mutants --in-place --in-diff` on this diff, at the 27.1.0 the workflow pins: 14 mutants, 11 caught, 3 unviable, 0 missed. `./scripts/test.sh` passes (224 browser tests, 0 failures). `cargo test` is green, 91 in the library and 258 across the integration suites, nine of them new here. `cargo fmt --check` and `cargo clippy --all-targets -- -D warnings` are clean. `check-gemini` was run against the live API, since this changes the setup message and that is a real regression risk.

Not verified end to end: an actual ten-minute close followed by a resume. That needs a real interview long enough to hit the cap. The wire format, the reader and the budget are covered by tests, but the reconnect itself has not been watched happening. A run past ten minutes should log `Gemini session resumed (1 so far)` and greet only once.

Reported as #2.

### DIFF
--- a/src/gemini.rs
+++ b/src/gemini.rs
@@ -1,4 +1,5 @@
 use std::io;
+use std::sync::{Arc, Mutex};
 use std::time::Duration;
 
 use base64::{Engine as _, engine::general_purpose::STANDARD};
@@ -15,6 +16,12 @@ use crate::runtime::{RuntimeBootstrap, TOOL_LOG_HINT, TOOL_READ_EDITOR};
 
 const LIVE_WEBSOCKET_ENDPOINT: &str = "wss://generativelanguage.googleapis.com/ws/google.ai.generativelanguage.v1beta.GenerativeService.BidiGenerateContent";
 const SETUP_TIMEOUT: Duration = Duration::from_secs(15);
+/// Bounds the socket itself, where `SETUP_TIMEOUT` bounds the exchange over it.
+/// Without this a connect that never answers hangs its caller for as long as
+/// the OS allows, and the caller is `run_room`'s loop: an interview whose
+/// candidate leaves during a stalled reconnect would not notice they had gone,
+/// and its own deadline would not fire either.
+const CONNECT_TIMEOUT: Duration = Duration::from_secs(15);
 /// Per attempt, so three tries plus backoff stay inside `REPORT_TIMEOUT`.
 const REPORT_ATTEMPT_TIMEOUT: Duration = Duration::from_secs(12);
 const REPORT_RETRY_BACKOFF: [Duration; 2] = [Duration::from_secs(1), Duration::from_secs(3)];
@@ -26,6 +33,12 @@ pub struct GeminiLiveSession {
     writer: SplitSink<WebSocketStream<MaybeTlsStream<TcpStream>>, Message>,
     reader: JoinHandle<()>,
     events: Receiver<GeminiEvent>,
+    /// Shared with the reader task, which is the only writer. Held beside the
+    /// event channel rather than sent through it because it is wanted at
+    /// exactly the moment the channel is finished: after the socket closed and
+    /// the last events were drained, when the caller has to decide whether it
+    /// can resume.
+    resumption: Arc<Mutex<Option<String>>>,
 }
 
 impl GeminiLiveSession {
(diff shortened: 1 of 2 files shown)


alanhc:
@cubic-dev-ai both findings were real and both are fixed in 60f860d.

P1 was the more serious of the two, and it meant this PR did not fix the bug it claims to. Every arm of `run_room` that writes to Gemini propagated its error with `?`. A socket the server has closed fails the next send well before `next_event` drains and returns `None`, and `pump_audio` flushes every hundred milliseconds, so the write side would win that race nearly every time: the interview ended from a dropped frame, and the arm that resumes the session never ran. The four writes (audio, video, the nudge, and the media attach) now log and continue, which leaves the close to the one arm that can explain it. Dropping a frame costs a tenth of a second of audio the resumed session did not need.

P2: `connect_async` had no timeout, where the exchange over it had `SETUP_TIMEOUT`. Added `CONNECT_TIMEOUT`, same 15s, so a connect that never answers cannot hold the loop that also notices the candidate leaving and fires the interview deadline. Attempts were already bounded at 16; that part stands, and `take_resume_attempt` now has tests on both sides of the boundary.

The other thing the first push got wrong was having no test for that logic at all. The mutation gate caught four survivors, every one of them in the function holding it. That logic is now `take_resume_attempt`, a pure function with three tests, and the diff runs 14 mutants with 0 missed.

jserv:

PR #4, opened 2026-08-25
Keep the lobby's start control on screen

alanhc:
The problem grid is 150 cards, which makes the lobby about 7500 pixels tall, and the duration row and "Start interview" sat under all of it at roughly 7400. On a 770 pixel viewport that is ten screens of scrolling past a wall of problems, with nothing above the fold saying there is anything below it, so the page reads as one that offers problems and no way to begin. The button was never disabled; it was unreachable.

A "launch-bar" wrapper holds both sections and sticks to the bottom of the viewport, so picking a problem and starting are visible together from the first screen. It takes the top margin the duration row carried and gives "start-row" a left auto margin, which puts the durations and the button at opposite ends of one bar instead of stacking them.

The bar also offers ten minutes. "MIN_DURATION_MIN" is ten, so it is the shortest interview the server will honour, and the lobby offered nothing below thirty. A shorter run is what trying the thing out wants, and on the free tiers it is the difference between a handful of attempts and a couple. Forty-five stays preselected: the test pinning the lobby's default to "DEFAULT_DURATION_MIN" is guarding a real invariant, since that constant is also what "DEFAULT_RECORDING_MAX_MINUTES" is defined as, and validation refuses a recording cap below the interview length.

## Verification

`./scripts/test.sh` passes, including
`browser_interview_duration_matches_the_server_clamp`, which is the test that
governs this area: it pins the lobby's preselected duration, its `app.js`
initial value, and the `interview.js` fallback to `DEFAULT_DURATION_MIN`, and
checks every offered button against the server's clamp. Ten minutes passes that
check because `MIN_DURATION_MIN` is ten; making it the default would not, which
is why forty-five stays selected.

Sticky positioning was confirmed in a browser at several scroll offsets: the bar
sits at the bottom of the viewport from the first screen down, and the start
button stays hit-testable throughout.

### DIFF
--- a/web/index.html
+++ b/web/index.html
@@ -629,15 +629,21 @@
         </button>
       </section>
 
-      <section class="duration-row" aria-label="Duration">
-        <button class="duration-button" type="button" data-duration="30">30 min</button>
-        <button class="duration-button selected" type="button" data-duration="45">45 min</button>
-        <button class="duration-button" type="button" data-duration="60">60 min</button>
-      </section>
+      <!-- Sticky: the problem grid is 150 cards tall, so a start control that
+           scrolls with it sits ten screens below the fold and reads as a page
+           with no way to begin. -->
+      <div class="launch-bar">
+        <section class="duration-row" aria-label="Duration">
+          <button class="duration-button" type="button" data-duration="10">10 min</button>
+          <button class="duration-button" type="button" data-duration="30">30 min</button>
+          <button class="duration-button selected" type="button" data-duration="45">45 min</button>
+          <button class="duration-button" type="button" data-duration="60">60 min</button>
+        </section>
 
-      <section class="start-row">
-        <button id="start" class="start-button" type="button">Start interview</button>
-      </section>
+        <section class="start-row">
+          <button id="start" class="start-button" type="button">Start interview</button>
+        </section>
+      </div>
 
       <section id="history" class="history" hidden></section>
 
--- a/web/styles.css
+++ b/web/styles.css
@@ -166,11 +166,24 @@ p {
   font-size: 0.875rem;
 }
 
+.launch-bar {
+  position: sticky;
+  bottom: 0;
+  z-index: 5;
+  display: flex;
+  flex-wrap: wrap;
+  align-items: center;
+  gap: 0.75rem;
+  margin-top: 1.5rem;
+  padding: 0.75rem 0;
+  border-top: 1px solid var(--hairline);
+  background: var(--page);
+}
+
 .duration-row {
   display: flex;
   flex-wrap: wrap;
   gap: 0.5rem;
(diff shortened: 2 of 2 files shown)


alanhc:
@cubic-dev-ai I don't think this one holds, and it is worth being concrete because the claim is about how `position: sticky` behaves.

A bottom-anchored sticky element is pinned to the bottom of the viewport whenever its natural flow position is below the viewport and its containing block still covers that area. The containing block here is `main`, which spans the whole page, so the bar is pinned from the very first screen rather than only once the candidate reaches it. That is the difference between `bottom: 0` and `top: 0` sticky, and it is why `fixed` is not needed.

Measured on this branch at `scrollY` 0, with a 748px viewport and a 7536px page: `.launch-bar` occupies viewport y 678 to 748, flush with the bottom of the first screen; `#start` is fully inside the viewport; and `document.elementFromPoint` at the button's centre returns `#start` itself, so it is hit-testable rather than merely painted.

Swept across the scroll range (0, 500, 2000, 4000, 6000, bottom), the button stays at viewport y 691 to 736 the whole way down, and lands at 556 to 601 at the end, where the bar reaches its natural position above the licence footer. Both `inViewport` and `clickable` are true at every one of those offsets.

`fixed` was the other option considered, and it was rejected because it would sit over the footer credit at the end of the page, which sticky avoids by design.

jserv:
Instead of displaying all LeetCode entries upfront, I plan to hide them initially and let users select their preferred difficulty level(s). The system will then randomly pick an entry from the selected level(s). Does this approach make more sense?

alanhc:
Yes — that makes more sense than what this PR does. This PR keeps the 150-card wall and makes the start control reachable in spite of it. Hiding the grid removes the reason the wall was a problem in the first place, and once the lobby is short the sticky bar stops earning its place.

One thing the redesign makes visible: difficulty and duration are two separate controls today, and they are not really independent. Ten minutes on a Hard problem is not a shorter interview, it is a failed one, and an hour on an Easy one is mostly waiting. If difficulty becomes an explicit choice, the two probably want to be coupled — picking a level sets a sensible default length the candidate can still override. Otherwise the lobby offers combinations no real interview would.

Worth noting the inverse framing too: a real interview fixes the slot first, and the interviewer picks something that fits it. So "choose how long you have, and we pick a problem for it" is arguably closer to the thing being simulated than choosing difficulty directly. Either way the two stop being independent.

Worth keeping alongside the random pick: a way to ask for a specific problem. Someone who stalled on Min Stack yesterday wants Min Stack today, not a random Medium.

On this PR: it should be closed if the redesign is near. The ten-minute option in it has the same problem I just described — it only really makes sense for Easy — so it may be better folded into the coupling above than kept as a standalone change.

jserv:
Close in favor of #15

Issue #1, opened 2026-08-24
程式碼編輯區問題：Tab 縮排、#include<stack>


nosba0957:
1. 在面試的程式碼撰寫頁面，沒辦法使用 Tab 縮排，會跳到下一個網頁元素。
2. 我在寫 min stack 題目時，繳交發現要自己動手寫 #include<stack>，否則會編譯錯誤。

alanhc:
The two halves of this turned out to have different answers.

The `#include` one is real, and #5 fixes it. `cppPrelude` in `web/compiler-explorer.js` listed seventeen headers and `<stack>` was not among them, so `stack<int>` failed to compile and typing the include by hand was the only way through. Reproduced against the compiler the runner actually asks for, gcc 16.2 at `-std=c++20`: `error: 'stack' does not name a type; did you mean 'obstack'?`. `<queue>` was already listed, which is what makes it look like an oversight rather than a decision.

Checking the rest of the bank found the same gap elsewhere: `<unordered_set>` was missing though `<set>` and `<unordered_map>` were there and five problems want it, and eight problems bound an answer with `INT_MAX` or `numeric_limits` with neither `<climits>` nor `<limits>` present. C has the identical defect for the identical reason, so those eight fail there too on `use of undeclared identifier 'INT_MAX'`. Both preludes are fixed together.

The Tab one appears to be already fixed on main, and predates this report: fe22b64 landed on 2026-08-22, two days before this issue was filed, and its commit message describes the same symptom. Since a fix landing before a report is not proof on its own, the current behaviour was checked in a browser against this tree rather than taken on trust. Tab inserts four spaces and calls `preventDefault`, Shift+Tab outdents, and Escape followed by Tab leaves the editor instead of indenting, which is the deliberate way out for anyone driving the page from the keyboard.

@nosba0957, that suggests you were on a build from before 2026-08-22. If you pull main and Tab still moves focus, please say so here, because then it is something other than what that commit fixed and worth reopening on its own.

PR #5, opened 2026-08-25
Give the C and C++ preludes their missing headers

alanhc:
A candidate solving Min Stack in C++ writes `stack<int>` and the compile fails with "'stack' does not name a type", because "cppPrelude" listed seventeen headers and `<stack>` was not among them. The only way through was to type `#include <stack>` at the top of the editor, which is not something a candidate should have to think about while an interviewer watches, and not something the starter code suggests.

`<queue>` was already there, which is what makes the omission look arbitrary rather than deliberate: the two are sibling container adaptors, and on libstdc++ including one does not declare the other. The same gap covers headers whose containers the bank's own problems want. `<set>`, `<map>` and `<unordered_map>` were listed and `<unordered_set>` was not, though five problems reach for it. Eight problems bound an answer with INT_MAX or numeric_limits, and neither `<climits>` nor `<limits>` was there. `<numeric>`, `<deque>`, `<array>` and `<tuple>` round out the set by the same rule: a container or algorithm a solution to something in this bank would naturally use.

C has the identical defect for the identical reason. Its prelude carried stdbool, stdio, stdlib, string and time, so the same eight problems fail on "use of undeclared identifier 'INT_MAX'". `<limits.h>` and `<math.h>` join it.

Java is unaffected: its harness imports "java.util.*", so a wildcard already covers what an explicit list has to keep up with.

Both preludes were checked against the compilers the runner actually asks for rather than a local toolchain, since the point is what godbolt builds: gcc 16.2 at "-std=c++20" and clang 19.1 at "-std=c17". Before, the C++ case failed as reported and the C case failed on INT_MAX. After, a MinStack using "stack", "unordered_set", "deque", "array", "accumulate", "numeric_limits", INT_MAX and "make_tuple" compiles and runs, as does a C solution using INT_MAX and "sqrt".

Reported as sysprog21/codetrial#1.

## Verification

Both preludes were compiled against the compilers `web/runners.js` actually
names, not a local toolchain: `g162` at `-O2 -std=c++20` and `cclang1910` at
`-O2 -std=c17`, through godbolt's own API.

Before, the reported case failed exactly as described:

```
error: 'stack' does not name a type; did you mean 'obstack'?
```

and C failed the same way on `use of undeclared identifier 'INT_MAX'`.

After, a MinStack using `stack`, plus `unordered_set`, `deque`, `array`,
`accumulate`, `numeric_limits`, `INT_MAX` and `make_tuple`, compiles and runs
with the expected output; so does a C solution using `INT_MAX` and `sqrt`.
`./scripts/test.sh` passes (224 browser tests, 0 failures).

## On the other half of the issue

sysprog21/codetrial#1 also reports that Tab moves focus out of the editor. That
one is already fixed on main, by fe22b64 on 2026-08-22, two days before the
issue was filed. Since a fix that predates a report is not evidence on its own,
the current behaviour was checked in a browser against this tree: Tab inserts
four spaces and calls `preventDefault`, Shift+Tab outdents, and Escape then Tab
leaves the editor rather than indenting. The reporter was most likely on a build
from before that commit.

### DIFF
--- a/web/compiler-explorer.js
+++ b/web/compiler-explorer.js
@@ -110,7 +110,9 @@ export function mapCompilerResponse(response) {
 }
 
 function cHarness(spec, candidateCode) {
-  return `#include <stdbool.h>
+  return `#include <limits.h>
+#include <math.h>
+#include <stdbool.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
@@ -234,20 +236,28 @@ ${spec.cases.map((testCase, index) => cppCase(spec, testCase, index)).join("\n")
 
 function cppPrelude() {
   return `#include <algorithm>
+#include <array>
 #include <chrono>
+#include <climits>
 #include <cmath>
+#include <deque>
 #include <functional>
 #include <iomanip>
 #include <iostream>
+#include <limits>
 #include <list>
 #include <map>
+#include <numeric>
 #include <optional>
 #include <queue>
 #include <set>
 #include <sstream>
+#include <stack>
 #include <stdexcept>
 #include <string>
+#include <tuple>
 #include <unordered_map>
+#include <unordered_set>
 #include <utility>
 #include <vector>
 using namespace std;


jserv:

PR #18, opened 2026-08-31
Run every interview as a real interview

jserv:
Runs every interview as a real interview, with REACTO and STAR as the
agent's guiding structure and scoring basis rather than as a mode the
candidate picks.

## What changed

Practice and scored modes are gone. `InterviewMode` and everything that
branched on it were deleted rather than deprecated, including the
practice-only prompt policy, the retry and drill controls, and the
lobby's mode selector. Reports written before the split still display
the mode they recorded; nothing invents one for a report without it.

REACTO and STAR are now presented as two separate scenario tables that
do not cross-reference each other, shown in a centered hint window that
dismisses itself before the interviewer speaks. Progress appears as a
compact checkmark list in the top-left corner, ticked from evidence the
interviewer actually banks, so the page never claims a phase the
transcript does not support.

## Fixes worth calling out

The report schema was rejected by Gemini on every request, so the branch
had never produced a finished report. Two separate defects: five
`additionalProperties` keys, which are not part of Gemini's `Schema`,
and an integer-valued `enum`, where the field is a repeated string. A
test now checks the schema against the keywords the API actually
accepts, so the next addition fails locally instead of at runtime.

The egress contract test described the request body by hand-copying
field names. It now decodes the body with the same generated type
LiveKit decodes it with, which immediately found two fields the
hand-written list had missed.

Grounding snippets are bounded by total bytes, not only per item, since
22 snippets of 240 multi-byte characters overshoot the prompt budget by
several times.

## Testing

`cargo clippy --all-targets` clean, 12 test binaries green, 342 browser
tests green. Fixes that could regress silently have a test that fails
when the fix is reverted.

## Note for review

Interview turns are occasionally cut short by the agent. The transcript
tail is now logged at the cut so the next run can distinguish a genuine
barge-in from acoustic echo. Not yet diagnosed, and not introduced by
this branch: the relevant code is byte-identical to main.

---
## Summary by cubic
Runs every interview as a real interview instead of a mode the candidate picks, with REACTO and STAR now driving the agent's structure and scoring. The practice and scored modes are gone, including their UI selectors and coaching controls; every interview now follows the same adaptive REACTO coding flow followed by a time-gated STAR behavioral round. The report contract is now enforceable end-to-end, and reports were previously rejected by Gemini on every request due to schema defects.

**Bug Fixes**
- Five invalid `additionalProperties` keys and an integer-valued `enum` in the report schema caused every report generation to fail; a test now checks the schema against the keywords Gemini actually accepts.
- The egress contract test now decodes the request body with the same generated type LiveKit uses, which caught two fields the hand-written list missed.
- Grounding snippets are bounded by total bytes, not just per item, since 22 multi-byte snippets overshot the prompt budget.
- Replay event ingestion now has a rate limit keyed on the account, and observation grants sweep on write instead of on the read path.
- Pause now stops the interview without extending the deadline, and it works after a disconnect; paused media no longer drops track-end frames or swallows the next reply.

**New Features**
- The lobby accepts an optional target role, seniority, and company profile plus a job description and resume, all validated and labeled untrusted in prompts.
- Progress is shown as a per-phase checkmark list ticked from evidence the interviewer banks, with REACTO and STAR shown as separate scenario cards.
- Reports now carry a versioned contract bundle, and rubric calibration fixtures gate scoring changes behind a release threshold; phase scores are marked formative and not for hiring decisions.

<sup>Written for commit b31d299dc6ad4cf561a8d630a738f6ae7d2893e4. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sysprog21/codetrial/pull/18?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


src/agent.rs
```diff
@@ -299,6 +856,10 @@ pub struct DataEventResult {
     pub update_last_interjection: bool,
     pub generate_reply: Option<String>,
     pub finish_interview: Option<String>,
+    /// Some only when a practice-mode pause state genuinely changed.
```
alanhc: Stale after this branch: the comment calls this "a practice-mode pause state", but practice mode is gone, and `apply_control` says so explicitly — "Pause used to be a practice-only affordance, and practice is gone. Left working for everyone rather than deleted with the mode."

As written the two disagree about whether the field still has a purpose, and this one reads as though `pause_changed` were left behind by the mode removal rather than deliberately kept.

src/web/mod.rs
```diff
@@ -394,8 +402,9 @@ impl AppState {
 /// supplies the implementation.
 ///
 /// The provider is handed over rather than looked up again from the room name.
-/// Re-deriving it is precisely the mistake [`room_and_provider`] documents: two
-/// pools assembled by two directory scans can disagree, and the candidate then
+/// Re-deriving it is precisely the mistake `pool::room_and_provider` documents:
+/// two pools assembled by two directory scans can disagree, and the candidate
+/// then
```
alanhc: The reflow after swapping the intra-doc link for `pool::room_and_provider` left `then` alone on its own line, so the paragraph renders with a three-word orphan: "...and the candidate / then / holds a token for one LiveKit project..."

alanhc (commented):
Read through the branch and ran the gate locally: `cargo test --locked` 485 passed / 0 failed / 2 ignored across 12 binaries, `cargo clippy --all-targets -- -D warnings` clean, browser suite 316 passed / 33 skipped / 0 failed.

Two small things below, both cosmetic — no correctness issues found.

The parts I looked at hardest held up. The profile and grounding sanitizers are candidate-controlled text reaching a prompt, and they escape with `{:?}`, label the block untrusted, and scope what the text may influence rather than trusting the model to infer it; bounding grounding on total bytes rather than per item is the right call given 22 items of 240 multi-byte characters. `RoomAuthorizations::allows` enforces expiry on the entry it reads instead of relying on the sweep, and the tests pin the boundary second in both directions. `jpeg_frame_geometry` bounding on bytes rather than pixels, with the reasoning written down, is a genuinely good catch. And `report_schema_uses_only_what_gemini_accepts` is honest about checking four rules rather than legality in general, which is worth more than a test that overclaims.

alanhc:
Re-ran the gate on `dc89bd5` (my earlier numbers were from `cadfb0b`, before the rebase): `scripts/test.sh` green end to end — 484 Rust tests passed / 0 failed / 2 ignored, browser 316 passed / 33 skipped / 0 failed, clippy and fmt clean. The count moved 485 → 484 because `dc89bd5` folds the two overlapping observer-grant tests into one, not because anything stopped running. Both notes above still apply at this head.

The three commits added since are worth calling out as good: `90fbc76` in particular fixes a real feedback loop — re-queueing the refused batch left the queue back over `REPLAY_MAX_BATCH`, so every subsequent event flushed on sight and earned another 429 for the rest of the window. Honouring `Retry-After` with a ceiling is the right shape.

On `52a48d2`: I went looking for a hole in the `Object.keys(item).length === knownEvidenceFields.size` shortcut — a row with seven keys where a known field is missing and an unknown one is present would take the fast path and drop the extra. It isn't reachable: every one of the seven known fields is required by the validation above it, so a row that gets that far with seven keys has exactly the known set. Noting it only because the shortcut's safety rests on that coupling, so a future optional field on an evidence row would quietly make it wrong.

PR #19, opened 2026-09-01
Stop offering C for class-style problems

jserv:
Seven of the 150 problems in the bank carry a class-style judge, and
`generateHarness` refuses a class judge in C. Nothing said so: C was one of
five tabs on every problem, fully stocked with C starter code, and the way a
candidate found out was by writing a solution and pressing run.

## What changed

`harnessGap(language, spec)` in `web/compiler-explorer.js` answers why a judge
cannot be run in a language, as a sentence a candidate can read, or null when
it can. `languagesFor` filters the tab list through it, and the interview page
disables from it and shows it as the tooltip, so a dead tab always states its
own reason and a second withheld pair is a second case in one function rather
than a second place to remember.

The judge is fetched without blocking the page. Every tab is offered until the
answer lands, which is what the row did before and what it keeps doing if the
judge never loads; the run path reports that failure itself. A candidate who
reached C in that window keeps their buffer and loses the runner, rather than
having their work replaced by a starter buffer at whatever moment the fetch
happened to land.

`updateRunAvailability` now owns the run button, asking `codingClosed()` rather
than inferring the interview phase from `nodes.editor.disabled`. That also
closes a hole where pausing and resuming handed the runner back on a language
the harness refuses.

`scripts/gen-wire-fixtures.mjs` used to learn the offered languages by
regex-scraping a `const languages = [...]` line out of `web/interview.js`. It
imports `ALL_LANGUAGES` now. The generated fixtures are byte-identical.

## Two independent commits ride along

`Stop the evidence carry from resting on a key count` removes a fast path in
`sanitizeReport` that was correct only because all seven known evidence fields
happen to be required; one optional field later and a row with an unknown key
would have dropped it.

`Take the rtrb fix for RUSTSEC-2026-0274` is a lockfile-only bump that was
failing the `cargo-audit` gate before this branch.

## Checks

`./scripts/test.sh` green end to end: 484 Rust tests passed / 0 failed / 2
ignored, browser 351 passed / 1 skipped / 0 failed, cargo-audit clean, fmt,
clippy at `-D warnings`, eslint and the generator checks all passing.

Verified in a real browser against a local release build, including the race
the non-blocking fetch opens:

| case | C tab | selected | run button |
| --- | --- | --- | --- |
| `lru-cache` (class judge) | disabled, reason in title | python | live |
| `two-sum` (function judge) | live, click selects it | c | live |
| `lru-cache`, judge stalled 2.5s, C grabbed first | live, then disabled | c, buffer kept | live, then dead |

---
## Summary by cubic
Stops offering C on the seven problems with class-style judges, so candidates no longer write a whole solution and press run to discover the C harness can't build it. The C tab now disables itself with the reason shown as its tooltip, and the judge is fetched without blocking the page: every tab stays offered until the response lands, and a candidate already in C keeps their buffer but loses the run button.

**Bug Fixes**
- `sanitizeReport` no longer assumes all evidence fields are required, so a future optional field with an unknown key won't be dropped.
- The run button now reads `codingClosed()` directly, closing a hole where pausing and resuming re-enabled the runner on an unsupported language.

**Dependencies**
- Bumps `rtrb` to 0.3.5, the lockfile-only fix for RUSTSEC-2026-0274 that the `cargo-audit` gate was failing on.

<sup>Written for commit 3fe99728d6d9a8b7fb773febda4af301c0e081ea. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sysprog21/codetrial/pull/19?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2153,9 +2153,9 @@ dependencies = [
 
 [[package]]
 name = "rtrb"
-version = "0.3.4"
+version = "0.3.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "4ade083ccbb4bf536df69d1f6432cc23deb7acccff86b183f3923a6fd56a1153"
+checksum = "fae8ee26b0371a29a77d2b2d6b3ae13aa81def6f9bf1b1b92a32d279a5e709b7"
 
 [[package]]
 name = "rusqlite"
--- a/scripts/gen-wire-fixtures.mjs
+++ b/scripts/gen-wire-fixtures.mjs
@@ -21,18 +21,7 @@ const ROOT = join(dirname(fileURLToPath(import.meta.url)), "..");
 const FIXTURES = join(ROOT, "tests", "fixtures");
 
 const lib = await import(join(ROOT, "web", "lib.js"));
-
-// Read the offered languages out of the producer instead of restating them.
-// The point of these fixtures is to catch the two sides disagreeing, and a
-// hardcoded list here would be a third place to disagree with. A language the
-// tabs offer but src/agent.rs does not recognize reaches no prompt and changes
-// no state, so the candidate's click is silently ignored.
-function offeredLanguages() {
-  const source = readFileSync(join(ROOT, "web", "interview.js"), "utf8");
-  const match = source.match(/^const languages = \[(.*?)\];$/m);
-  if (!match) throw new Error("could not find the language list in web/interview.js");
-  return match[1].split(",").map((part) => part.trim().replace(/^"|"$/g, ""));
-}
+const { ALL_LANGUAGES } = await import(join(ROOT, "web", "compiler-explorer.js"));
 
 const CODE = "def two_sum(nums, target):\n    return []\n";
 const EDITED = "def two_sum(nums, target):\n    seen = {}\n    return []\n";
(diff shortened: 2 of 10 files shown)


alanhc (approved):
Looks good — no findings. Ran `./scripts/test.sh` on `3fe9972`: green end to end, 484 Rust passed / 0 failed / 2 ignored, browser 0 failed. (My browser split reads 319 passed / 33 skipped against your 351 / 1; both total 352, so it's how node tallies rather than a coverage difference — I'm on v25.4.0.)

The shape is right. `harnessGap` returning the sentence rather than a boolean is what makes the tab and its tooltip incapable of disagreeing, and `generateHarness` still throwing independently means the rule is enforced even if a caller forgets to ask. Defaulting to `languagesFor(null)` so an unreadable judge offers everything is the correct direction to fail: withholding a language over a dropped request would be a worse guess than letting the run path report it.

The race is handled where I'd have worried about it. `applyLanguages` never calls `setLanguage`, so a late judge cannot swap a candidate's buffer out from under them, and routing the run button through `updateRunAvailability` means the C-selected case loses the runner rather than the work. Ordering the `then` after `bindEvents` with the microtask reasoning written down is a nice touch — that's the kind of thing that silently rots otherwise.

On `f39db54`: this is a better fix than the coupling I pointed at on #18. Dropping the key-count shortcut removes the dependency on all seven fields staying required, and the null prototype catches something I had missed entirely — a `__proto__` key would have hit the inherited setter and vanished. I checked it directly: the field round-trips as an own property and `Object.prototype` is untouched. Building the carry only when there is something to carry also makes the common row cheaper than the version it replaces, so nothing was traded for the correctness.

`22cac4b` is scoped the way a lockfile bump should be, and swapping the regex scrape in `gen-wire-fixtures.mjs` for a real import removes a third place for the language list to disagree — `readFileSync` is still used further down, so no dead import left behind.

PR #20, opened 2026-09-01
Make the recording checks prove what they claim

jserv:
The recording provisioning and delivery checks reported success without
observing what they claimed. This makes each of them prove its own rule, and
adds the tests that fail when a rule is removed.

## The recording checks

`scripts/recording-provision-check.sh` proved the delivery account could reach
the bucket and the Shared Drive, which is exactly what a far too powerful
account proves too. It now reads the policies as a separate auditor account,
walks the whole ancestor chain, and refuses the three shapes where exact access
is unreadable rather than guessing: a role reaching the account through
`allUsers`, a group, a domain or a principal set; a conditional binding; and a
bucket without Uniform Bucket-Level Access, where per-object ACLs grant access
no IAM policy shows.

Two reads in that script could never have worked against real `gcloud`. The
bucket resource is fetched with `--raw` now, because the checks name fields the
way the JSON API names them and the standardised form renames them, which made
the lifecycle refusal fire on correctly configured buckets. The ancestor walk
refused any entry without a `resource` field, which that command does not emit,
and which was only ever used to name the ancestor in a message.

`scripts/recording-integration.sh` gains its delivery and cleanup phases, which
previously printed a note and exited 2, leaving the deletion path with no
end-to-end proof. The reader permission is judged against the moment it is read
and required to still be live then: the server sets that expiry from its own
clock after upload, so comparing it against Drive's `createdTime` would have
rejected every correct delivery, and accepting a lapsed one would have called
dead access a successful handover.

## Also here

- `src/agent.rs` was 2154 lines holding six concerns, past what its own module
  doc claimed. It is now `agent/{events,report,value}.rs` beside the existing
  submodules. Everything is re-exported, so no caller changed and all 69 items
  are the same code in a different file.
- `scripts/check-todo-deps.py` is deleted. It was a gate in `scripts/test.sh`
  that had been green without reading anything, looking for section headings
  that no longer exist.

## Verification

`./scripts/test.sh` exits 0: 489 Rust tests across 12 binaries, 351 browser
tests. Every commit was checked out and its suites run, so the branch bisects
clean.

Every new refusal was mutation tested: the check is removed, the suite must
fail, the check is restored. That found several rules already in the tree with
no test that could fail, which are now covered.

## Known limits

The `gcloud` field names are read from Google's documentation, not observed,
because the development checkout has no `gcloud`. The test stub now answers the
way the tool does, so dropping `--raw` fails the suite, but that pins the flag
rather than the names. The first credentialed run is what confirms them, and it
is what Recording 0 exists for.

---
## Summary by cubic
Makes the recording provisioning and delivery checks verify what they claim, instead of reporting success without observing it.

- The provision check now audits IAM policies with a separate auditor account, walks the whole ancestor chain, and refuses access shapes it can't verify exactly: roles through `allUsers`, a group, a domain or a principal set; conditional bindings; and buckets without Uniform Bucket-Level Access.
- Fixed two reads that could never have worked against real `gcloud`: the bucket is fetched `--raw` now, and the ancestor walk no longer refuses entries without a `resource` field.
- The integration harness runs the delivery and cleanup phases it previously skipped (it printed a note and exited 2), judging the reader permission against the moment it is read so lapsed access fails rather than passing as a successful handover.
- Operators must supply two new variables, `CODETRIAL_RECORDING_PROJECT` and `CODETRIAL_RECORDING_IAM_AUDITOR_JSON`, or the provision check refuses to run.
- Split `src/agent.rs` into `agent/{events,report,value}.rs` with everything re-exported, and deleted `scripts/check-todo-deps.py`, a gate that had been green without reading anything.
- `gcloud` field names are pinned from Google's documentation, not observed, since the checkout has no `gcloud`; the test stub matches the tool, so the first credentialed run confirms them.

<sup>Written for commit 7cbeb1039b22347319681a76e8e9ca6b1f2710ed. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sysprog21/codetrial/pull/20?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


alanhc:
Read the branch at `ff38165` against `main` at `d18f985`. Nothing below was executed: this is a source read, and where a claim rests on documented API behaviour rather than a run I say so. The six cubic findings all look genuinely addressed on the branch, under different SHAs than the bot's links after the rebase (`399bcc19`, `6c21c81a`, `0fc46d78`), so this is what is left.

### The Shared Drive audit reads only the first page

`scripts/recording-provision-check.sh:178` requests

```
files/$drive_id/permissions?supportsAllDrives=true&fields=permissions(id,type,role,emailAddress,deleted)
```

with no `pageSize` and no `nextPageToken`. Because `fields` names only `permissions(...)`, the token is excluded from the response, so truncation is not merely unhandled but invisible. Per the Drive v3 documentation `permissions.list` returns at most 100 results when `pageSize` is unset *for files in a shared drive*; the unpaginated behaviour is the non-shared-drive case, and `$drive_id` here is a shared drive.

The conclusion drawn from it is an exactness claim:

```python
if len(matches) != 1 or matches[0].get("type") != "user" or matches[0].get("role") != "organizer":
    raise SystemExit(f"{delivery} must have exactly one active Shared Drive organizer permission")
```

On a Drive with more than 100 members, a second permission for the delivery account landing on page 2 leaves `matches` at length one, and the audit reports exactly-one while more exist. The other direction is safe: a sole permission on page 2 makes the check refuse. But this direction is the script asserting an exactness it did not observe, which is the shape this PR is removing everywhere else.

Nothing in the suite can catch it. `tests/test_recording_provision_check.py` stubs `curl` with a fixed one-entry `drive.json`, so page size is unmodelled.

`scripts/recording-integration.sh:397` has the same shape, but against a freshly created file's permissions, where a hundred entries is not a real case. Worth matching for symmetry rather than for risk.

### Bare `nervous` is still accepted, while bare `anxious` is refused

`src/agent/report.rs:250-268` gained `" nervousness "`, `" nervously "`, `" nervous demeanor "` and the six `appeared` / `appears` / `seemed` / `looked` / `sounded` / `come(s) across as nervous` forms, but not `" nervous "` itself. `" anxious "` and `" anxiety "` are both in the list bare.

Against the normalizer at `:209`, which turns every non-alphanumeric into a space and pads the result:

- `The candidate was nervous throughout` passes
- `The candidate was anxious throughout` is refused

Same judgment, opposite outcomes. And since the match is a contiguous substring, `appeared very nervous` also passes: one adverb defeats all six enumerated forms.

`confident` staying out of the list reads as deliberate, and I would leave it: *confident in their approach* is a legitimate technical observation. `nervous` has no such use, and `anxious`, `personality` and `charisma` already set the precedent for the bare token.

### `member_roles` may refuse every real organization

`scripts/recording-provision-check.sh:200` refuses on any `allUsers`, `allAuthenticatedUsers`, `group:`, `domain:` or `principalSet:` binding found anywhere in the ancestor chain, including bindings with no relation to the delivery account. The reasoning in the comment is right — you cannot prove the account is outside a group without a directory lookup — but a real organization essentially always carries at least one `group:` binding at the org or folder level, so the likely outcome is a check that refuses on correct setups rather than on misconfigurations, and a provisioning gate nobody can pass proves as little as one that passes without looking.

The fixtures do not settle it either way: `ancestors_iam` defaults to a single project entry with `"bindings": []`. Either the refusal wants to tell the operator what to do next — audit the group and re-run with it attested — or the check wants the directory lookup. Raising it as a decision rather than a defect.

### Checked and fine

- **The agent split is faithful.** I extracted every top-level `fn` / `struct` / `enum` / `const` / `static` / `trait` / `type` from `src/agent.rs` at `d18f985`, and from `agent.rs` plus `agent/{events,report,value}.rs` on the branch: 97 items on each side, set difference empty in both directions.
- **The new `config.rs` comment's claim holds.** It names the provisioning check as the authoritative DNS gate, and the script does resolve the host and refuse any answer that is not `is_global` at `:100-108`, before either credential is written to disk.
- **`62d65530f` is equivalent rather than an off-by-one.** Exactly one `push` precedes the length check, so the buffer can exceed `MAX_INTEGRITY_EVENTS` by at most one and `remove(0)` does what `drain(0..excess)` did, without the unsigned subtraction.

Issue #16, opened 2026-08-31
A 60-minute interview loses its recording at minute 50


alanhc:
Picking 60 minutes in the lobby starts an interview whose recording is reaped ten minutes before it ends.

`token_duration_min` (src/web/token.rs:190) clamps the requested length to `MIN_DURATION_MIN..MAX_DURATION_MIN`, that is 10..90 (src/config.rs:55-56), so 60 passes through untouched. The recording cap is a separate, deployment-wide number: `max_minutes` is read from `CODETRIAL_RECORDING_MAX_MINUTES` once at startup (src/config.rs:937) and defaults to `DEFAULT_RECORDING_MAX_MINUTES` = `DEFAULT_DURATION_MIN` = 45 (src/config.rs:835, :50). `stale_after` for a recording in progress is `max_minutes * 60 + 300` (src/recording/sweeper.rs:334, :331), which is 50 minutes on that default, and `sweep` fails anything older as `LostWebhook` (:554). Nothing compares the two numbers.

The config check that looks like it covers this, `max_minutes < interview_duration_min` (src/config.rs:943), compares against `CODETRIAL_DURATION_MIN` rather than the length of the interview being started, and the lobby offers 60 whatever that is set to. The `const _: () = assert!` added in #15 pins the two defaults against future drift, but is satisfied by definition today since `DEFAULT_RECORDING_MAX_MINUTES` is `DEFAULT_DURATION_MIN`.

`suggestedDuration` in web/app.js already declines to suggest 60 for this reason. The button stays.

The smallest fix is probably to bound the duration against the recording cap in `token_duration_min`, where it is already clamped, so the sweeper enforces the same number the interview was given. Making `max_minutes` per-recording would also work; refusing it only in the lobby would leave a hand-crafted request able to do the same thing.

Note: this is from reading `main` at 2463634, not from watching a recording get reaped.

jserv:
Close via #17

PR #21, opened 2026-09-01
Tell the interview page how long its interview is

alanhc:
`/api/token` clamps the requested length to the range and, where this server records, to the recording cap (#16, #17), and hands that number to the agent in the room metadata. The page never saw it. `durationMin` in web/interview.js came from the query string, clamped only to `MIN_DURATION_MIN..MAX_DURATION_MIN`, and the response body was `token`, `serverUrl` and `roomName` and nothing else — so the browser had no way to learn what it had been granted short of opening a signed token it has no business opening.

Three things read that number, and all three could describe an interview nobody was having:

- the countdown, whose deadline the server builds from the metadata value instead (`boot.duration_min`, src/livekit.rs:408);
- the coding and behavioral split, which decides when the page announces the round transition that the agent gates on its own `coding_minutes`;
- the entry written to the report history when it is over.

The response now carries the granted length and the page applies it before it joins the room, which is also before the replay's opening lifecycle event writes the round split down. An answer that is missing or unusable leaves the requested length in place, so a page served by an older build behaves exactly as it did.

The clamp on the query string stays as the fallback for a URL that did not come through the lobby, and `browser_interview_duration_matches_the_server_clamp` still pins its bounds to the server constants. That test cannot reach the other clamp: the recording cap is per deployment, so there is no constant to pin it to, and being told is the only way the page can know it. Its own comment is what pointed at this — *"the browser countdown and `run_room`'s server-side hard deadline are both built from this number"* — and the cap is the half of that number the test cannot see.

`TokenResponse` drops `Eq` to carry a `serde_json::Value`, because a JSON number can be a float. Nothing compares the type, and the alternative is rounding the one value that has to reach the page exactly as the agent reads it.

Tests: the endpoint's body and its token metadata must name the same length; `token_response` reports what it granted with no cap, under a cap, and for a fractional length; and the page is checked for reading `connection.durationMin` before `connectLiveKit`. `token_api_matches_frontend_contract_over_http` pins the response's whole key set, so it failed on the new field until it was updated — which is the test doing its job.

`./scripts/test.sh` exits 0, with the browser tier actually running: 352 node tests, 351 pass, 0 fail, 1 skip (the Java harness, no `javac` here), and no Playwright skips. cargo-audit, actionlint and shellcheck are skipped locally for want of the tools and run in CI.

---
## Summary by cubic
Tells the interview page the interview length it was actually granted, so the countdown, round split, and saved report no longer describe an interview nobody is having. `/api/token` always clamped the length (to the range and, where this server records, the recording cap) and handed it to the agent, but the page only ever saw the requested query-string value.

- `/api/token` now returns `durationMin` in the response body.
- The page applies it before joining the room, before anything reads the length.
- A missing or unusable value leaves the requested length in place, so pages served by an older build behave exactly as before.
- `TokenResponse` drops `Eq` to carry a `serde_json::Value`, since a JSON number can be a float.

<sup>Written for commit b1069d9ae6f6a11c34aeebcaca8b4e90f2b4bc4c. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sysprog21/codetrial/pull/21?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/web/token.rs
+++ b/src/web/token.rs
@@ -46,11 +46,21 @@ pub struct TokenConfig<'a> {
 /// credential a holder can join a room with, not a key that merely mints them.
 /// It is the shortest-lived secret here and the easiest to print by accident,
 /// because it looks like a response body rather than a key.
-#[derive(Clone, PartialEq, Eq)]
+///
+/// No `Eq`, because `duration_min` is a JSON number and one of those can be a
+/// float. Nothing compares these, and the alternative is rounding the value
+/// that has to reach the page exactly as the agent reads it.
+#[derive(Clone, PartialEq)]
 pub struct TokenResponse {
     pub token: String,
     pub server_url: String,
     pub room_name: String,
+    /// The length this interview actually got, which is not always the length
+    /// that was asked for: `token_duration_min` clamps to the range and, where
+    /// this server records, to the recording cap. The page runs its own
+    /// countdown and its own round split, so it has to be told the answer
+    /// rather than keep the question it asked.
+    pub duration_min: Value,
 }
 
 /// `/api/token` mints a real LiveKit credential and is necessarily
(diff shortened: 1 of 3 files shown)


jserv:

PR #23, opened 2026-09-04
Make a double-clicked binary able to start

nosba0957:
A release binary had no way to enter credentials: the only input path was
hand-editing `codetrial.env.local` before ever running it. A double-clicked
exe on Windows exits before anyone can read why — no arguments, no config,
and a console that closes with the process.

`web` mode now starts anyway. With no config anywhere it serves a Setup page,
validates each value against the live service (LiveKit through `ListRooms`,
Google through a real Gemini Live handshake), writes `codetrial.env.local`,
then rebinds as the ordinary app in the same process. Zero arguments now mean
`web`; startup prints the URL; a failed start writes `codetrial-error.log`.
Both files, and the config search, sit beside the executable, which is the one
thing about a double-click that does not depend on where it was started from.

An operator deployment is untouched: `--config` still decides, a named but
missing one is still an error, and `GOOGLE_API_KEY` stays optional.

13 new tests in `tests/cli.rs` (18 → 31), driving the real binary. Every commit
builds and passes on its own.

The first commit is unrelated: the `pre-commit` hook checked out only the staged
diff, leaving a module parent without its children and failing rustfmt over
untouched files. It is first because the rest of the branch trips over it.

---
## Summary by cubic
Lets a double-clicked Windows binary start: previously the only input path was hand-editing a config file, and the exe exited before any error was visible. `web` mode with no config anywhere now serves a Setup page that validates credentials, writes `codetrial.env.local`, and rebinds as the app; operator deployments are unaffected.

- Zero arguments now defaults to `web` mode; startup prints the URL to open.
- The Setup page validates LiveKit and optional Google credentials against the live services before writing the config file beside the executable.
- The config search now also checks beside the executable, after `./config/` and the working directory.
- Any web startup failure writes `codetrial-error.log` beside the executable.
- Fixes the pre-commit hook checking out only the staged diff, which broke rustfmt over untouched files.

<sup>Written for commit 97b86d2d9881747014dddfd969a082572f92843e. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sysprog21/codetrial/pull/23?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -232,8 +232,11 @@ fn run_web(options: CliOptions) -> Result<(), String> {
     // Falls through: once `serve_setup` returns, the config exists and the rest
     // of this function is an ordinary launch, on the socket Setup served on.
     // Why that socket is carried here rather than rebound: `serve_setup`.
-    let handed_over = if is_cold_start(&options) {
-        Some(serve_setup(&options)?)
+    // Read once, so the question and the page it opens answer from the same
+    // map.
+    let before_config = environment_and_flags(&options);
+    let handed_over = if is_cold_start(&options, &before_config) {
+        Some(serve_setup(&before_config)?)
     } else {
         None
     };
@@ -408,16 +411,42 @@ fn web_provider_pool(
     Ok(pool)
 }
 
-/// The solo self-serve cold start: no `--config` given and none of the
-/// searched paths holds a file. A named-but-missing `--config` is an
-/// operator's mistake, not this.
+/// The solo self-serve cold start: no `--config` given, none of the searched
+/// paths holds a file, and the environment has not supplied the credentials
+/// either. A named-but-missing `--config` is an operator's mistake, not this.
 ///
 /// The search order comes from `primary_config_path` rather than being
 /// written out again here. Spelled twice it was free to drift, and the drift
 /// is silent in the worst direction: a Setup page in front of someone whose
 /// config the rest of the process is about to read.
-fn is_cold_start(options: &CliOptions) -> bool {
-    options.config_path.is_none() && primary_config_path(options).is_err()
+///
+/// The environment counts because a file is not the only way to be configured.
+/// A headless launch with `LIVEKIT_*` exported and no file used to exit naming
+/// the file it wanted, which is a failure someone reading a log can act on;
+/// asking it instead put a page nobody would open in front of a process that
+/// then waited forever. Only the LiveKit keys, because those are what the web
+/// side refuses to start without, and `GOOGLE_API_KEY` decides whether this
+/// process also hosts interviewers rather than whether it can run.
+fn is_cold_start(options: &CliOptions, values: &BTreeMap<String, String>) -> bool {
+    options.config_path.is_none()
+        && primary_config_path(options).is_err()
+        && !environment_supplies_credentials(values)
+}
+
+/// Whether this launch is already configured without a file.
+///
+/// Its own function so it can be asserted without a filesystem: `is_cold_start`
+/// reaches this only after the config search misses, and the search hits in any
+/// checkout that has a config, which is every machine a developer runs the
+/// suite on.
+///
+/// The LiveKit keys and not `GOOGLE_API_KEY`, because those three are what the
+/// web side refuses to start without. The Google key decides whether this
+/// process also hosts interviewers, which is a shape rather than a requirement.
+fn environment_supplies_credentials(values: &BTreeMap<String, String>) -> bool {
+    ["LIVEKIT_URL", "LIVEKIT_API_KEY", "LIVEKIT_API_SECRET"]
+        .iter()
+        .all(|key| nonempty(values, key).is_some())
 }
 
 /// Serves Setup until a submission writes `codetrial.env.local`, then returns
@@ -429,14 +458,8 @@ fn is_cold_start(options: &CliOptions) -> bool {
 /// connections this page just served, and `TIME_WAIT` outlasts any retry worth
 /// writing. The loser has already written the config, so it has also spent the
 /// cold start that would have brought Setup back.
-fn serve_setup(options: &CliOptions) -> Result<std::net::TcpListener, String> {
-    // The environment and the flags, which in a cold start is everything
-    // `load_values` has: the file it reads on top of them is the file that does
-    // not exist yet.
-    let mut values = std::env::vars().collect::<BTreeMap<_, _>>();
-    apply_options(&mut values, options);
-
-    let listener = bind_web_listener(&value_or(&values, "CODETRIAL_WEB_ADDR", DEFAULT_WEB_ADDR))?;
+fn serve_setup(values: &BTreeMap<String, String>) -> Result<std::net::TcpListener, String> {
+    let listener = bind_web_listener(&value_or(values, "CODETRIAL_WEB_ADDR", DEFAULT_WEB_ADDR))?;
 
     // Fails closed on an address the socket cannot name, for the reason
     // `run_web` does: an impossible kernel answer must not be the way past a
@@ -444,13 +467,13 @@ fn serve_setup(options: &CliOptions) -> Result<std::net::TcpListener, String> {
     let bound = listener
         .local_addr()
         .map_err(|error| format!("bound listener has no address to check: {error}"))?;
-    if let Some(refusal) = public_setup_refusal(&values, bound) {
+    if let Some(refusal) = public_setup_refusal(values, bound) {
         return Err(refusal);
     }
     // Everything `run_web` will refuse for that this already knows the answer
     // to. There is one such rule today; the rest of its checks need the pool
     // the submission has not supplied yet.
-    if let Some(refusal) = production_secret_refusal(&values) {
+    if let Some(refusal) = production_secret_refusal(values) {
         return Err(refusal);
     }
     // The window stays open now, but still needs to say where to go.
@@ -474,7 +497,7 @@ fn serve_setup(options: &CliOptions) -> Result<std::net::TcpListener, String> {
             let listener = tokio::net::TcpListener::from_std(listener)?;
             let service = codetrial::web::setup_service(
                 ready.clone(),
-                is_production(&values),
+                is_production(values),
                 bound.port(),
                 cold_start_config_path(),
             );
@@ -648,6 +671,20 @@ fn load_agent_config(options: &CliOptions) -> Result<AgentConfig, String> {
         &provider_order,
     );
     Ok(config)
+}
+
+/// The environment and the flags, which in a cold start is everything
+/// `load_values` has: the file it reads between them is the file that does not
+/// exist yet.
+///
+/// Its own function so `run_web` does not name the process environment. Reading
+/// a deployment key straight from it there is what
+/// `binary_web_reads_deployment_keys_from_the_config_file` refuses, because a
+/// key set in the config file was once silently a no-op.
+fn environment_and_flags(options: &CliOptions) -> BTreeMap<String, String> {
+    let mut values = std::env::vars().collect::<BTreeMap<_, _>>();
+    apply_options(&mut values, options);
+    values
 }
 
 fn load_values(options: &CliOptions) -> Result<BTreeMap<String, String>, String> {
@@ -1391,15 +1428,52 @@ mod tests {
             ..super::CliOptions::default()
         };
 
+        let empty = values(&[]);
         assert!(
-            !super::is_cold_start(&named(root.join("Cargo.toml"))),
+            !super::is_cold_start(&named(root.join("Cargo.toml")), &empty),
             "a config file that exists was named, so there is nothing to set up"
         );
         assert!(
-            !super::is_cold_start(&named(root.join("no-such-config.env"))),
+            !super::is_cold_start(&named(root.join("no-such-config.env")), &empty),
             "a config file that is missing was still named, and a name is an \
              instruction to read that file rather than an invitation to ask"
         );
+    }
+
+    /// Exported credentials are a configured launch, so there is nothing to ask
+    /// for. Asking anyway put a page nobody would open in front of a headless
+    /// start that then waited forever, where naming the missing file and
+    /// exiting was something a log could carry.
+    #[test]
+    fn exported_credentials_are_a_configured_launch() {
+        assert!(super::environment_supplies_credentials(&values(&[
+            ("LIVEKIT_URL", "wss://example"),
+            ("LIVEKIT_API_KEY", "key"),
+            ("LIVEKIT_API_SECRET", "secret"),
+        ])));
+
+        // One short of the set the web side refuses to start without, which is
+        // still a launch that has to be asked.
+        assert!(!super::environment_supplies_credentials(&values(&[
+            ("LIVEKIT_URL", "wss://example"),
+            ("LIVEKIT_API_KEY", "key"),
+        ])));
+
+        // Present and blank is missing, the way `nonempty` reads it everywhere
+        // else, and the way `read_config_file` would have written it back.
+        assert!(!super::environment_supplies_credentials(&values(&[
+            ("LIVEKIT_URL", "wss://example"),
+            ("LIVEKIT_API_KEY", "key"),
+            ("LIVEKIT_API_SECRET", "   "),
+        ])));
+
+        // The optional one is not part of the question.
+        assert!(super::environment_supplies_credentials(&values(&[
+            ("LIVEKIT_URL", "wss://example"),
+            ("LIVEKIT_API_KEY", "key"),
+            ("LIVEKIT_API_SECRET", "secret"),
+            ("GOOGLE_API_KEY", ""),
+        ])));
     }
 
     #[test]
--- a/src/main.rs
+++ b/src/main.rs
@@ -215,6 +215,11 @@ fn bind_web_listener(addr: &str) -> Result<std::net::TcpListener, String> {
 }
 
 fn run_web(options: CliOptions) -> Result<(), String> {
+    if is_cold_start(&options) {
+        // Falls through: once `serve_setup` returns, the config exists, so
+        // the rest of this function is an ordinary launch — same process.
+        serve_setup(&options)?;
+    }
     let values = load_values(&options)?;
 
     // The built-in default is a published string, so in production it is not a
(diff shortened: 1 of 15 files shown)


jserv (changes requested):
Resolve regressions, rebase latest `main` branch, resolve conflicts, and response to cubic AI reviews.

src/main.rs
```diff
@@ -214,6 +219,11 @@ fn bind_web_listener(addr: &str) -> Result<std::net::TcpListener, String> {
 }
 
 fn run_web(options: CliOptions) -> Result<(), String> {
+    if is_cold_start(&options) {
+        // Falls through: once `serve_setup` returns, the config exists, so the
+        // rest of this function is an ordinary launch — same process.
+        serve_setup(&options)?;
```
jserv: The setup listener is dropped here and `run_web` binds the same address again a few lines down. On Windows, `std::net::TcpListener::bind` deliberately does not set `SO_REUSEADDR` (it allows socket hijacking there), so the connections this page just served leave the port in TIME_WAIT and the rebind can fail with WSAEADDRINUSE. That is the one platform this feature exists for, and the failure mode is a saved config plus an exit, with no cold start left to retry from. Either keep the bound socket and swap the router on it, or retry the bind for a couple of seconds before giving up.

src/web/setup.rs
```diff
@@ -0,0 +1,287 @@
...
+
+    // From `pairs`, so the file holds exactly the config that was checked and
+    // probed above. Every value is known to carry no newline by now, which is
+    // what makes one line per key a faithful encoding of it.
+    let contents = pairs
+        .iter()
+        .map(|(key, value)| format!("{key}={value}\n"))
+        .collect::<String>();
+    let path = crate::exe_dir().join("codetrial.env.local");
+    if let Err(error) = std::fs::write(&path, contents) {
```
jserv: `fs::write` creates the file with the process umask, so `LIVEKIT_API_SECRET` and `GOOGLE_API_KEY` land at 0644 and every local user can read them. It also follows a pre-existing symlink. Open it with `OpenOptions::new().write(true).create_new(true)` plus `.mode(0o600)` under `#[cfg(unix)]`, and report the collision instead of truncating.

src/web/setup.rs
```diff
@@ -0,0 +1,287 @@
...
+use crate::runtime::bootstrap;
+
+/// Notified on a successful submission: the caller's shutdown signal to
+/// drop this listener and rebind as the full app.
+///
+/// `production` is the caller's, not one read here. The launch this hands over
+/// to enforces the production rules on the file this writes, and a second
+/// reading of `NODE_ENV` is a second answer waiting to disagree with it.
+pub fn setup_service(ready: Arc<Notify>, production: bool) -> Router {
+    Router::new()
```
jserv, line 44: Nothing checks the `Host` header, so the loopback bind is not the confinement it looks like. A page on any origin can point a short-TTL DNS name at 127.0.0.1, rebind after the first load, and POST here as same origin, which skips the preflight that `Content-Type: application/json` would otherwise force. The payload is the attacker's own LiveKit project, written into the config every later interview routes through. Reject any request whose `Host` is not `localhost`, `127.0.0.1` or `[::1]` with the bound port.

src/main.rs
```diff
@@ -506,9 +631,16 @@ fn primary_config_path(options: &CliOptions) -> Result<PathBuf, String> {
...
+
+    // The working directory first, which is a checkout's `config/` and an
+    // operator's deployment directory, then the folder the executable sits in.
+    // A release binary is unpacked into a folder of its own and its config is
+    // written there, so it has to be findable from a shortcut or a terminal
+    // opened somewhere else, not only from a double-click.
     for path in [
         PathBuf::from(DEFAULT_CONFIG_PATH),
         PathBuf::from("codetrial.env.local"),
+        codetrial::exe_dir().join("codetrial.env.local"),
```
jserv: Config and the error log now resolve against the executable, but `DEFAULT_ACCOUNT_DB_PATH` is still a bare `codetrial.db`, so the accounts database resolves against the working directory. Launched from a shortcut or a terminal elsewhere, the process finds the config it wrote and then creates a fresh empty database beside wherever it was started, losing accounts between launches. It bites harder on this path: if that directory is not writable, `initialize_accounts` fails after `codetrial.env.local` is already on disk, so the next launch is no longer a cold start and the Setup page never comes back. Default the database beside the executable too.

src/web/setup.rs
```diff
@@ -0,0 +1,287 @@
...
+        // Same proof as `check-gemini`: open a real session.
+        // `CODETRIAL_GEMINI_LIVE_URL` lets tests redirect this away from the
+        // real endpoint.
+        let room_name = format!("{}-smoke", config.room_prefix);
+        let boot = bootstrap(&config, &room_name, None, config.default_duration_min);
+        let url = std::env::var("CODETRIAL_GEMINI_LIVE_URL")
+            .unwrap_or_else(|_| gemini_live_websocket_url(&config.google_api_key));
+        match open_live_session_at(&url, &boot, None).await {
+            Ok(session) => {
+                let _ = session.close().await;
```
jserv, line 385: `GeminiLiveSession::shutdown` is `self.writer.close().await?; self.reader.abort();`, so a close that errors returns early and never aborts the reader. The `let _` here then drops the `JoinHandle` and the task keeps the socket alive for the rest of the process. Abort the reader before propagating the close result.

src/main.rs
nosba0957: I would choose to put both files under `<root>/config/`: `codetrial.env.local` and `codetrial.db` together. `<root>` is the checkout for a source build, and the folder holding the binary for a downloaded one. The database would go in the same directory as the config file that was actually loaded, which is the directory `provider_dir` already resolves for provider files. The config search order itself would not change.
My worry with defaulting it beside the executable is that there are two install shapes. A downloaded binary sits in a folder the user chose, which is stable. `make` puts the binary in `target/`, which `make clean` deletes. Beside the executable is right for the first, but would lose the database on the second.

What do you think?

jserv: Agree. Place these configurations in `config/`.

alanhc (commented):
Reviewed at 9857a59. The cold-start path is well tested and the
`livekit_scheme` consolidation is a real fix, but a few paths let the Setup
page take credentials and then strand the process, and the page itself is
reachable in a way the loopback bind does not cover.

## Blocking

**No `Host` check on the Setup server (`src/web/setup.rs:24`).** The router
is `/healthz`, `/`, `POST /api/setup` with no middleware, and nothing reads
`Host`, `Origin`, or `Referer`. `public_setup_refusal` gates the bind
address, which DNS rebinding does not have to defeat: a page the user visits
while Setup is open rebinds its own domain to 127.0.0.1:3000, becomes
same-origin, and can POST `application/json` with no preflight. The
attacker's own valid LiveKit credentials pass validation, get written, and
the app rebinds against the attacker's project. The `Json` extractor blocks
form-encoded CSRF but not this. Rejecting a request whose `Host` is not the
bound address closes it.

**A production cold start wedges after a successful save
(`src/main.rs:232`).** With `NODE_ENV=production` and no `SESSION_SECRET`,
`public_setup_refusal` still serves the page, the submission is probed,
`codetrial.env.local` is written, 200 is returned, and then `run_web` hits
the SESSION_SECRET refusal and exits. The browser polls for 10s and reloads
into a dead port. Because the file now exists, `is_cold_start` is false on
every later launch, so the page never comes back and the reason is only in
`codetrial-error.log`. The comment at `setup.rs:209` scopes this to the URL
rule, but the SESSION_SECRET verdict is computable from the map
`serve_setup` already builds, so it can refuse before taking credentials
rather than after writing them.

**An empty optional Google field shadows an env key
(`src/web/setup.rs:273`).** `pairs` always contains `GOOGLE_API_KEY`, so a
blank field writes a literal `GOOGLE_API_KEY=` line. `read_config_file` does
not skip it (only blank and `#` lines are skipped), and `load_values` inserts
file pairs over the environment, so `load_from_pairs` reports the key missing
and the process degrades to web-only. Before this change a file without the
key let the environment supply it. The `eprintln!` that says so goes to the
console a double-click does not have. Omitting the line when the field is
empty is enough.

**Submitted values are used untrimmed (`src/web/setup.rs:220`).** The
required-field loop trims, and its comment explains why, but nothing after it
does:

- `" wss://host"` (a leading space survives most pastes) passes
  `validate_livekit_url`, which trims internally, then reaches
  `livekit_http_base`, where `livekit_scheme` only matches at offset 0. The
  URL comes back unrewritten, reqwest refuses the `wss` scheme, and working
  credentials are reported as "did not work".
- A key or secret with a trailing space is signed into the JWT verbatim and
  401s, though the same value hand-written into `codetrial.env.local` works,
  because `read_config_file` trims on read.
- A whitespace-only `googleApiKey` fails the untrimmed `is_empty()` at
  `setup.rs:237`, enters the Gemini branch, and surfaces "missing required
  config keys: GOOGLE_API_KEY" for a field the form calls optional.

One trim at the top of `submit_setup` covers all three.

## Worth fixing

**Port 0 rebinds somewhere else (`src/main.rs:425`).** `serve_setup` never
passes its bound address forward, so `run_web` re-resolves the same string at
`main.rs:292`. With `--web-addr 127.0.0.1:0`, a form `main.rs:956` tests, the
kernel hands out a different port and `waitForRestart` polls the old one for
10s before reloading into a dead socket. Passing the listener's `local_addr`
forward also gives the retry the doc comment at `main.rs:414` promises;
`bind_web_listener` currently has none, which is the same gap that loses a
fixed port to TIME_WAIT on the platform this PR targets.

**`is_cold_start` ignores credentials in the environment
(`src/main.rs:409`).** A loopback-bound headless launch with `LIVEKIT_*`
exported and no file used to exit 1 with "required configuration file is
missing"; it now blocks indefinitely on a page nobody will open. Non-loopback
binds still fail fast, so this is confined to the default a naive unit file
uses, but a fast exit was the more diagnosable failure.

**`a_spawned_server_pipes_stderr_so_a_bind_failure_can_be_read`
(`tests/cli.rs:1316`).** The comment says "it exits on a usage error
immediately", which this PR made false. The test bypasses `cli_command`, so
nothing scrubs the environment and the cwd is the crate root: in a checkout
with `config/codetrial.env.local` it boots a full server, and a lost bind
race writes `codetrial-error.log` next to the real test binary, which is the
contamination `binary_beside` exists to prevent. Spawning an unknown mode
keeps the fast exit the assertion wants.

**Quote-edged values change between probe and reload
(`src/web/setup.rs:276`).** `read_config_file` does
`trim().trim_matches('"').trim_matches('\'')`, and the values are written
raw, so a secret whose first or last character is a quote is probed as one
string and loaded as another. Low reachability, since a fully quote-wrapped
paste fails the probe first and self-corrects, but it contradicts the "file
holds exactly the config that was checked and probed" comment.

## Smaller

- `public_setup_refusal` (`src/main.rs:472`) spells
  `published_secret_refusal`'s predicate again rather than sharing it. The
  doc comment says the test is that function's; a third reachability signal
  added to one and not the other drifts silently, and in the direction that
  leaves the credential-writing page open.
- `validate_livekit_credentials` (`src/livekit.rs:936`) rebuilds the Twirp
  POST that `livekit_room_service_request` already makes, and drops the
  response body from the error. That body carries LiveKit's own reason, which
  the Setup form is the one place a user needs. Only the token differs and
  both are strings, so a shared core taking (base, token, method, body)
  covers both.
- The Gemini probe error at `src/web/setup.rs:262` is formatted raw, while
  every other caller of this chain wraps it in `redact_api_key`
  (`main.rs:514`, `main.rs:534`, `livekit.rs:1788`, `gemini.rs:256`). The
  close-frame reason is folded in at `gemini.rs:611`, so server-controlled
  text reaches the response unredacted.
- Six added comments use an em dash, in `src/main.rs`, `src/web/setup.rs`,
  and `tests/cli.rs`. `codetrial-conventions` asks for ASCII in source
  comments.

## Checked against 9857a59

Cold start in an empty directory, bound to 127.0.0.1:39117. A request
carrying `Host: evil.example.com` is served the Setup page with 200, and
`POST /api/setup` with the same header reaches the handler, answering
`{"error":"livekitUrl is required"}` rather than being rejected at the edge.
That is the validation error, so the endpoint is reachable under any `Host`
a rebound name supplies.

Three throwaway tests, all passing:

- `validate_livekit_url(" wss://example.livekit.cloud", false)` is `Ok`,
  `livekit_http_base` returns it unchanged, and the resulting `send()` fails
  with a reqwest builder error naming the surviving `wss://` URL. Without
  the space the same input rewrites to `https://`.
- The file `submit_setup` writes for a blank Google field reads back as a
  present, empty `GOOGLE_API_KEY`, which overwrites an environment value in
  the map `load_values` builds; `load_from_pairs` then reports the key
  missing.
- `LIVEKIT_API_SECRET='abc` written raw reads back as `abc`.

src/main.rs
alanhc: `submit_setup` writes with `std::fs::write`, which does not create the parent. A downloaded binary sits in a folder with no `config/` in it, so the first submission fails with `NotFound` on the one install shape the page exists for. The move needs a `create_dir_all` on the directory before the write, and that directory is then also what `provider_dir` resolves.

src/web/setup.rs
```diff
@@ -0,0 +1,287 @@
...
+    ] {
+        if field(key).chars().any(char::is_control) {
+            return super::json_response(
+                StatusCode::BAD_REQUEST,
+                json!({ "error": format!("{key} must not contain control characters") }),
+            );
+        }
+    }
+
+    let livekit_url = field("livekitUrl");
```
alanhc, line 305: The values that get validated and probed are not the values that get read back. `field` does not trim, the emptiness check trims a copy, and `read_config_file` trims what it parses, so a submission with surrounding whitespace is judged in one form and stored in another.

Paste ` wss://foo.livekit.cloud` with the leading space a copy usually brings. `validate_livekit_url` trims internally and accepts it. `livekit_http_base` then compares a fixed-length prefix, ` wss:/` matches no entry in `LIVEKIT_SCHEMES`, no rewrite happens, and the request goes out over `wss://`, which reqwest refuses before it leaves the process. The page answers "livekitUrl, livekitApiKey or livekitApiSecret did not work", naming credentials that are correct -- and the same string, once written and read back trimmed, is the one that would have worked.

Trim each field once, where `field` is defined, and let everything downstream see only the trimmed value.

src/web/setup.rs
```diff
@@ -0,0 +1,287 @@
...
+
+  <label for="livekitApiKey">LiveKit API Key</label>
+  <input id="livekitApiKey" name="livekitApiKey" type="text" required>
+
+  <label for="livekitApiSecret">LiveKit API Secret</label>
+  <input id="livekitApiSecret" name="livekitApiSecret" type="password" required>
+  <small>Both come from the same project's Keys page.</small>
+
+  <label for="googleApiKey">Google API Key</label>
+  <input id="googleApiKey" name="googleApiKey" type="password">
```
alanhc, line 145: Blank is accepted here and the resulting server looks finished: a lobby that loads, an interview that starts, and nothing that ever joins the room. The one place that says so is `run_web`'s `eprintln!` -- "Interviews will wait forever unless an agent joins each room from elsewhere" -- on the stderr of a process this branch exists to run without a console.

The field can stay optional, but the page has to carry that sentence next to it, or the app has to say it somewhere a browser can show.

src/web/setup.rs
```diff
@@ -0,0 +1,287 @@
...
+
+/// Notified on a successful submission: the caller's shutdown signal to
+/// drop this listener and rebind as the full app.
+///
+/// `production` is the caller's, not one read here. The launch this hands over
+/// to enforces the production rules on the file this writes, and a second
+/// reading of `NODE_ENV` is a second answer waiting to disagree with it.
+pub fn setup_service(ready: Arc<Notify>, production: bool) -> Router {
+    Router::new()
+        .route("/healthz", get(|| async { "ok\n" }))
```
alanhc: Nothing reads this now. The page polls `/api/session` precisely because `/healthz` cannot tell the two servers apart, and no other caller is left.

What it still does is answer a probe. A container `HEALTHCHECK` or a local supervisor gets 200 from a process that has no config, is serving a credential form, and will never serve an interview. Dropping the route removes that; answering with a status that names the setup state would too.

src/main.rs
```diff
@@ -386,6 +398,109 @@ fn web_provider_pool(
...
+
+/// Any `web` failure, not just a cold-start one. No attempt to tell a solo
+/// user's config apart from an operator's — the error text already says
+/// what's wrong. Best-effort: must not shadow the real error.
+///
+/// Beside the executable, where the config it just failed to read also
+/// lives. A double-clicked binary has no console to leave the reason in and
+/// no working directory anyone chose, so the folder it was unpacked into is
+/// the one place its owner knows to look.
+fn log_web_error(error: &str) {
```
alanhc, line 593: Written on every failure and removed on none, so the file outlives what it describes. The reader double-clicks, fails, finds `codetrial-error.log` beside the executable, fixes the cause, and the next launch succeeds with the file still sitting there. The launch after that goes wrong for a different reason and they read a message from days ago, in the one folder the error text sends them to.

Remove it once `run_web` is past its startup checks.

alanhc (commented):
Four comments inline. One more has no line in the diff to hang on:

`tests/cli.rs`, `a_spawned_server_pipes_stderr_so_a_bind_failure_can_be_read`: its comment says "No arguments: it exits on a usage error immediately", and this branch makes that false. Zero positionals now mean `web`, so the spawn is a cold start that binds the default `127.0.0.1:3000` and serves the Setup page until `ServerProcess::drop` kills it. It is also the only spawning test that inherits the environment whole -- no `env_remove` -- so an exported `LIVEKIT_*` set turns it into a full launch that writes `codetrial.db` into the repository root during `cargo test`. An unknown mode word keeps the assertion and restores the immediate exit.

nosba0957:
All threads are addressed. @jserv's five are in `6f5bd86` and `8072545`. @alanhc 's are in `8216857` (blocking), `361624e` (worth fixing), `ca0a786` (smaller) and `95081f4` (the three later threads). Except the `Host` check and the port-0 rebind, which were already fixed in `6f5bd86`, above the commit that review read.

alanhc (approved):
All four inline comments and the `create_dir_all` reply are addressed. Read at 649d3f8:

- `field` trims once at its definition, so validation, the probes and the file all see the same value.
- The blank Google field carries its consequence on the form, and an empty key is now left out of the file rather than written as an empty pair. The second half is a real find I missed: `read_config_file` keeps an empty value and `load_values` lays file pairs over the environment, so the line would have erased an exported `GOOGLE_API_KEY`.
- `/healthz` is gone from the Setup router; the full app still routes it.
- `clear_web_error_log` runs past the bind and `published_secret_refusal`, which is the first moment that can tell a fixed launch from a fresh failure.
- `a_spawned_server_pipes_stderr_so_a_bind_failure_can_be_read` spawns an unknown mode, and the comment matches what it does again.

The `try_clone` handover in 6f5bd86 is a better answer than the retry loop that thread was heading toward: the retained descriptor keeps the port bound across the swap, so there is no window to lose rather than a shorter one.

`361624e` corrects something I got wrong. I read `load_values` as always requiring a file and concluded a headless launch with `LIVEKIT_*` exported could not regress. It could: the old exit named the file it wanted, and the cold start replaced that with a page nobody would open.

Built and ran the suite locally on macOS at 649d3f8, all green: lib 151, main 20, agent 94, cli 44, config 36, recording 53 + 9 + 10, runtime 7, token 8, web 107 (2 ignored). One first-run failure in `vendored_assets_are_served_typed_and_cached` was a cold checkout racing `build.rs`'s vendor fetch, not this branch; it passes once the assets are on disk.

jserv:
Thank @nosba0957 for contributing!

PR #37, opened 2026-09-11
Allow LiveKit regional CSP hosts

### LINKED ISSUE #36: CSP blocks connections to LiveKit regional endpoints
sation-xxx.livekit.cloud` to a regional endpoint such as: `conversation-xxx.otokyo1b.production.livekit.cloud` 
However, the generated Content Security Policy only allows the original hostname. The browser therefore blocks 
the regional endpoint.
## Steps to reproduce
1. Configure CodeTrial to use a LiveKit Cloud project. 2. Run the server with `make serve`. 3. Open 
`http://localhost:3000`. 4. Start an interview. 5. Check the browser console.
## Actual result
The browser reports: `Refused to connect because it violates the document's Content Security Policy.` The 
LiveKit client then fails to establish the signal connection.
## Expected result
The browser should allow LiveKit regional endpoints and successfully establish the signaling connection.
## Investigation
Allowing HTTPS and WSS connections to `*.livekit.cloud` in `connect-src` removes the CSP error. After this 
change, I encountered HTTP 429 responses from LiveKit. That appears to be a separate service-side or 
account-related issue and is not caused by audio playback.
## Environment
- Windows with WSL - Browser opened at `http://localhost:3000` - CodeTrial built and served from WSL
- LiveKit Cloud

jserv:
LiveKit Cloud can redirect a browser to a regional signaling host after the configured endpoint. Keep the corresponding Cloud wildcard beside the exact HTTP and WebSocket sources, while rejecting endpoint spellings that cannot safely appear in a CSP.

Closes #36

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -506,22 +506,42 @@ pub(crate) fn livekit_scheme(url: &str) -> Option<&'static (&'static str, bool, 
     })
 }
 
-/// Rejects what `url_origin` in `web.rs` would otherwise have to cope with: a
-/// scheme nothing here knows, or a scheme with no host after it. The URL never
-/// appears in the message, because the message reaches stderr and a LiveKit URL
-/// can carry a query string.
+pub(crate) fn livekit_host_is_csp_safe(authority: &str) -> bool {
+    let Some((host, port)) = split_authority(authority) else {
+        return false;
+    };
+    let host_is_safe = if authority.starts_with('[') {
+        host.parse::<std::net::Ipv6Addr>().is_ok()
+    } else {
+        !host.is_empty()
+            && host
+                .bytes()
+                .all(|byte| byte.is_ascii_alphanumeric() || matches!(byte, b'.' | b'-'))
+    };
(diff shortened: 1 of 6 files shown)


alanhc:
Tests pass locally (205 unit tests).

The substantive change here is `csp_origin`, not the wildcard: the previous path went through `url_origin` alone, which only checks for printable ASCII, so a configured `wss://host;script-src=*` reached the header verbatim and the semicolon terminated `connect-src` early. Rejecting the endpoint and withholding its Cloud wildcard is the right call, and the test pins it.

Two notes:

- The host-character check would sit better in `validate_livekit_url`. As it stands a malformed host still starts the server and fails only in the browser, which is the shape of failure the rest of this file argues against.
- The `_ => return vec![origin]` arm in `livekit_origins_for_csp_origin` is now unreachable: every endpoint passes `validate_livekit_url`, so the scheme is one of the four.

alanhc (approved):

Patriciaath999 (approved):
I reviewed the changes. The CSP now allows LiveKit regional signaling hosts while validating unsafe endpoint formats.

PR #39, opened 2026-09-14
Keep grounding choices when the other file loads

ColtenOuO:
### Sumarry

In the lobby's document grounding section, every JD or resume upload rebuilt all three snippet lists from scratch. Checking a few JD requirements and then uploading a resume silently unchecked them, and the same happened in reverse.

#### Upload and select checkboxes

image

#### After uploading the resume

image

### Fix

Before a file is parsed, the current selection is read and only the groups that file replaces are dropped (`requirements` for a JD, `skills` and `anchors` for a resume), since those indexes point into a list that no longer exists. The rest is restored when the lists are re-rendered.

- `web/document-grounding.js`: add `retainedSelection(selected, kind)`
- `web/app.js`: extract `checkedGrounding()` (shared with the Start handler) and pass the retained selection to `renderGroundingChoices`

---
## Summary by cubic
Fixes the lobby's document grounding so uploading a resume no longer clears JD requirement checkboxes, and vice versa.

- Only the groups the new file replaces lose their selections; other checked groups survive the re-render.
- Extracts `checkedGrounding()` so the Start handler and file loading share the same selection logic.

<sup>Written for commit f328c7bb513ca6487b34d0543e822d5c4107e938. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sysprog21/codetrial/pull/39?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


web/app.js
```diff
@@ -228,10 +228,16 @@ async function loadGroundingFile(kind) {
     else { grounding.skills = []; grounding.anchors = []; }
     status.textContent = error.message;
   }
-  renderGroundingChoices();
+  renderGroundingChoices(selected);
```
jserv: `checkedGrounding()` runs before the `await`, so this restores what was checked when the file was picked rather than when the parse landed: check a JD requirement, pick a resume off a slow volume, uncheck that requirement while it reads, and it comes back checked. Nothing between the `await` and this call rebuilds the list, so reading the selection here instead is equivalent in every other case and closes that window (the earlier `const selected` then goes away):

```suggestion
  renderGroundingChoices(retainedSelection(checkedGrounding(), kind));
```

A read superseded by a second `change` on the same input, or by `Clear document grounding`, still commits out of order. That one predates this change and wants a per-kind generation checked before the commit, as a follow-up.

tests/browser/document-grounding.test.js
```diff
@@ -149,3 +149,11 @@ test("selection indexes outside the extracted list are dropped, not clamped", as
   // Order follows the indexes as given, not the document.
   assert.deepEqual(pickWith([2, 0]), ["Must have Go", "Must have Rust"]);
 });
+
+test("re-reading one document keeps the selection made in the other", () => {
```
jserv, line 153: This pins `retainedSelection` but not the regression: delete `checkbox.checked = selected[group].includes(index)` from `renderGroundingChoices` and this still passes while the unchecking comes straight back. `tests/browser/lobby.test.js` drives the real lobby, and its `lobbyTest` helper with `setInputFiles` covers it in a few lines: load a JD, check a requirement, load a resume, assert the JD box is still checked. That file's own header makes the point that this class of bug is invisible to a test that does not drive the page.

jserv (changes requested):
Rebase latest `main` branch and resolve conflicts.

tests/browser/document-grounding.test.js
ColtenOuO: Added a `lobbyTest` in `tests/browser/lobby.test.js`: load a JD, check the second requirement, load a resume, assert the JD boxes read `[false, true]`. Deleting `checkbox.checked = selected[group].includes(index)` makes it fail (expected true, got false); restoring it passes.

web/app.js
ColtenOuO: Thanks! I missed that detail.

The selection is now read after the parse lands, and the earlier `const selected` is gone. I'll leave the superseded-read race for a follow-up PR with a per-kind generation check.

ColtenOuO:
Thanks for reviewing!

Addressed all feedback from the inline comments and resolved merge conflicts.

jserv:

PR #40, opened 2026-09-14
Drop a grounding read that a newer one replaced

ColtenOuO:
Follow-up to #39.

### Sumarry
`loadGroundingFile` wrote its result whenever the file read finished, even if a newer action had happened in the meantime:

- **Second file on the same input**: pick a slow JD, then a second JD before the first finishes. The slow read lands last and replaces the　second file's snippets, so the input shows one file and the list　another. If the slow read fails, it clears the good result instead.
- **Clear during a read**: press Clear document grounding while a resume is still reading. When the read lands, the cleared snippets and the "Parsed locally" status come back.

### Fix and Change
Each document kind keeps a read counter (`groundingReads.jd` / `.resume`). A read takes the next number before its `await` and gives up afterwards if the number is no longer current, so it touches neither `grounding`, the status line, nor the lists. Clear bumps both counters. The counters are per kind so a JD read never cancels a resume read.

### New test

Two `lobbyTest`s in `tests/browser/lobby.test.js`. They hold the read of a file named `slow.txt` open by wrapping `Blob.prototype.arrayBuffer`, and release it on purpose.

---
## Summary by cubic
Fixes a race condition where a grounding file read that finished later could replace the result of a newer read, leaving the file list and snippets out of sync. Reads now use per-kind counters, and a read that is no longer the latest is dropped without touching the lists or status line.

**Bug Fixes**
- Picking a second JD before the first finishes reading no longer lets the slow read overwrite the second file's snippets.
- Clearing grounding while a resume is still reading no longer brings back the cleared snippets and "Parsed locally" status.
- Adds two lobby tests that hold a slow read open to cover both stale-read cases.

<sup>Written for commit 0f79ad5588d7b3478e926e93014fdd7695b1b8bb. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sysprog21/codetrial/pull/40?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


web/app.js
```diff
@@ -259,17 +260,19 @@ start.addEventListener("click", async () => {
...
-    status.textContent = "Parsed locally. Select only snippets you want to send.";
+    parsed = await parseGroundingFile(input.files[0], kind);
   } catch (error) {
-    if (kind === "jd") grounding.requirements = [];
-    else { grounding.skills = []; grounding.anchors = []; }
-    status.textContent = error.message;
+    message = error.message;
   }
+  if (read !== groundingReads[kind]) return;
+  if (kind === "jd") grounding.requirements = parsed?.requirements ?? [];
```
jserv, line 273: This write is the one that lands mid-flight, and the start handler is the other reader of `grounding` the counter does not cover: it captures `selected` from `checkedGrounding()` before the sign-in round trip but calls `selectedGroundingPacket(grounding, selected, consented)` after it. Pick a JD, check a requirement, pick a second JD, then press Start on a gated lobby: the second read lands during `recordGitHubLogin`, the stale indices now point into the new file's requirements, and the packet ships snippets the candidate never checked. Follow-up worth doing: snapshot the three grounding arrays next to `selected`, before the await, the way the comment above the handler says the rest of the form is read.

tests/browser/lobby.test.js
```diff
@@ -394,6 +394,61 @@ lobbyTest("loading a resume keeps the JD requirements already checked", async (p
...
+      if (this.name === "slow.txt") {
+        await gate;
+        done = read.call(this);
+        return done;
+      }
+      return read.call(this);
+    };
+    window.releaseSlowRead = async () => {
+      open();
+      while (!done) await new Promise((resolve) => setTimeout(resolve, 0));
```
jserv: This spins forever when the gated `arrayBuffer` was never reached, which is exactly what a regression in the `change` handler these tests exist to catch looks like. `page.evaluate` carries no timeout and `scripts/test.sh` runs `node --test` without `--test-timeout`, so the failure mode is the default gate hanging behind a busy page instead of a test that fails and names what was wrong. Bound the wait and throw.

```suggestion
      for (let wait = 0; !done && wait < 1000; wait++) await new Promise((resolve) => setTimeout(resolve, 10));
      if (!done) throw new Error("the gated read never started");
```

ColtenOuO: Yeah,  I missed this case. Addressed in lastest commit.

web/app.js
ColtenOuO: The bug is real and I think this solution is a great idea. To keep this PR within scope, I opened a separate PR to address it (#41).

Thanks for reviewing and point out this follow-up.

ColtenOuO:
The `check` failure here is not from this change. The only failing gate is `cargo-audit`, which flags `rustls` 0.23.43 under [RUSTSEC-2026-0285](https://rustsec.org/advisories/RUSTSEC-2026-0285) (TLS 1.3 handshake messages accepted across encryption level boundaries), published 2026-09-14. The advisory landed between runs: this branch passed in [run 34843141378](https://github.com/sysprog21/codetrial/actions/runs/34843141378) and failed in [run 34874694806](https://github.com/sysprog21/codetrial/actions/runs/34874694806) the same day.

The fix is `rustls` >= 0.23.45, in #42 . I will rebase this branch once that merges.

jserv:
> The fix is `rustls` >= 0.23.45, in #42 . I will rebase this branch once that merges.

Go ahead.

ColtenOuO:
rebase to the main, CI looks green!

web/app.js
jserv, line 273: This guards `grounding` against a stale read, but the start handler still reads it across an await: `selected` from `checkedGrounding()` and `consented` are captured before `await recordGitHubLogin(false)`, while `selectedGroundingPacket(grounding, selected, consented)` runs after it. Nothing disables the file inputs or the consent box during that round trip, so a JD picked while login is pending lands here and leaves the old indices in `selected` pointing at snippets the candidate never checked, and a Clear followed by a new pick ships them under a `consented` captured before the box was unchecked. A follow-up should build the packet before the await, the rule the handler already applies to `duration` and `problem`.

ColtenOuO: Moved into #41: the packet is now built before recordGitHubLogin, with the rest of the form, and stored only after sign-in answers. A start the packet refuses no longer spends a login.

PR #41, opened 2026-09-14
Ship the grounding snippets checked at start

ColtenOuO:
### Summary

Follow-up to the review on #40. On a lobby that requires sign-in, the start handler read the checked snippet indexes before `/api/login` but looked up their text after it. A JD picked during that wait replaced the lists, so the packet sent snippets the candidate never checked. This predates #39 and #40.

### Changes
- `web/app.js`: copy the three grounding lists next to `selected` before the sign-in await, and build the packet from that copy.
- `tests/browser/lobby.test.js`: hold `/api/login`, load a second JD during the hold, and assert the stored packet still has the first JD's snippet. Fails on `main` (ships `Must know SQL`), passes with the fix.

---
## Summary by cubic
Builds the grounding packet when Start is pressed, before the sign-in round trip, so a JD chosen during login cannot replace the snippets the candidate checked. Previously, the handler looked up snippet text after login and could store the wrong selections; it now stores the packet only after sign-in succeeds, rechecks consent at that point, and rejects invalid starts before signing in.

- Adds browser coverage for a JD change during login, consent withdrawal during login, and a refused start.

<sup>Written for commit 813b14680421fefb0ba05bec3f633e8b0da3595c. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sysprog21/codetrial/pull/41?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


ColtenOuO:
Waiting for #40 to be merged without conflicts, then I'll mark this ready for review.

ColtenOuO:
The `check` failure here is not from this change. The only failing gate is `cargo-audit`, which flags `rustls` 0.23.43 under [RUSTSEC-2026-0285](https://rustsec.org/advisories/RUSTSEC-2026-0285) (TLS 1.3 handshake messages accepted across encryption level boundaries), published 2026-09-14. The advisory landed between runs: this branch passed in [run 34843141378](https://github.com/sysprog21/codetrial/actions/runs/34843141378) and failed in [run 34874694806](https://github.com/sysprog21/codetrial/actions/runs/34874694806) the same day.

The fix is `rustls` >= 0.23.45, in #42 . I will rebase this branch once that merges.

jserv:
> The fix is `rustls` >= 0.23.45, in #42 . I will rebase this branch once that merges.

Go ahead.

ColtenOuO:
Fixed the implementation according to the review on #40.

tests/browser/lobby.test.js
```diff
@@ -917,6 +917,49 @@ lobbyTest("a start ships the length that was on screen when it was pressed", asy
...
+
+  const finishLogin = (() => {
+    let done;
+    holdLogin = new Promise((resolve) => (done = resolve));
+    return done;
+  })();
+  await page.fill("#github-login", "candidate");
+  await page.click("#start");
+  await page.click("#grounding-clear");
+  await page.uncheck("#grounding-consent");
```
jserv: `clearGrounding` already sets `groundingConsent.checked` to `false`, and `uncheck` returns immediately on a box that is already unchecked, so this line does nothing. It reads as coverage for revoking consent mid-start, but the only assertion here is on the snippet text, so that path is untested. Drop it, or make it a separate test that asserts a revoked consent leaves nothing in `codetrial.interview-grounding.v1`.

web/app.js
```diff
@@ -220,8 +220,13 @@ start.addEventListener("click", async () => {
   if (profile.seniority) destination.searchParams.set("seniority", profile.seniority);
   if (profile.targetCompany) destination.searchParams.set("company", profile.targetCompany);
   const focus = nodes.practiceFocusShareInput.checked ? practiceFocus(reports) : null;
-  const selected = checkedGrounding();
-  const consented = nodes.groundingConsent.checked;
+  let packet;
+  try {
+    packet = selectedGroundingPacket(grounding, checkedGrounding(), nodes.groundingConsent.checked);
```
jserv, line 226: Snapshotting consent here along with the snippets changes what `Clear document grounding` does during the sign-in wait: on `main` the clear won that race, since it empties `grounding`, `pick` drops the stale indexes, `selectedGroundingPacket` returns `null`, and nothing is stored. Now the pre-click packet is stored once the login resolves even though the candidate unchecked consent and cleared the lists, which is the one control whose later state should still win over the snapshot. Re-read `nodes.groundingConsent.checked` next to `storeGroundingPacket` and store `null` when it is no longer set.

ColtenOuO: Fixed in 813b146. `nodes.groundingConsent.checked` is read again right before `storeGroundingPacket`, and the packet becomes `null` when consent is no longer set, so clearing or unchecking during the sign-in wait stores nothing, as on `main`. Covered by the new test `consent withdrawn during the sign-in wait stores no grounding`, which fails without the fix.

tests/browser/lobby.test.js
ColtenOuO: Dropped the no-op `uncheck`, and split consent into its own test as suggested: `consent withdrawn during the sign-in wait stores no grounding` unchecks consent alone (no Clear) during the held login and asserts `codetrial.interview-grounding.v1` is `null`.

One deviation: with the consent fix, Clear now leaves nothing stored, so the snapshot test re-checks `#grounding-consent` after picking the second file. Without that line it fails with `undefined` instead of `["Must know Rust"]`, since it is testing that the pre-click snippets survive, not consent.

jserv:

PR #46, opened 2026-09-15
Preserve WebRTC linkage in macOS releases

### LINKED ISSUE #45: LiveKit runner crashes on macOS 26.2 arm64
## Description

When running CodeTrial on macOS 26.2 arm64, the Rust LiveKit runner crashes immediately during startup.

The browser shows `Mic on`, but the interview agent remains at `Waiting`. Running the LiveKit runner directly reproduces the same issue.

Gemini itself works correctly with `check-gemini`, so the failure appears to occur specifically in the LiveKit/WebRTC path before the interview agent can start processing audio.

## Error Logs

Running:

```bash
./codetrial-aarch64-apple-darwin run-livekit interview-test
```

results in:

```text
provider pooling: 1 provider(s) over 1 project(s): primary=wss://codetrial-13pd7gob.livekit.cloud

*** Terminating app due to uncaught exception 'NSInvalidArgumentException',
reason: '+[NSString stringForAbslStringView:]: unrecognized selector sent to class ...'

libc++abi: terminating due to uncaught exception of type NSException
```

The process exits with `ABRT`.

As a control, Gemini setup succeeds:

```bash
./codetrial-aarch64-apple-darwin check-gemini
```

and returns:

```text
Gemini setupComplete: model=gemini-3.1-flash-live-preview voice=Puck problem=two-sum duration=45min
```

## Expected vs Actual Behavior

- **Expected Behavior:** The LiveKit runner should successfully connect to the interview room, start the interview agent, and receive microphone audio from the browser.
- **Actual Behavior:** The LiveKit runner crashes immediately with `+[NSString stringForAbslStringView:]: unrecognized selector`, causing the agent to never start and the browser UI to remain at `Waiting`.

## Environment

- **OS:** macOS 26.2 (arm64)
- **Access Method:** Web browser via local server (`http://127.0.0.1:3000`)

abnormal749:
## Summary

Fix the macOS arm64 release crashing during WebRTC initialization when starting an interview:

```text
+[NSString stringForAbslStringView:]: unrecognized selector
```

The release workflow sets `RUSTFLAGS` to an empty string on macOS. This overrides the Apple-specific flags in `.cargo/config.toml`, dropping `-ObjC` and allowing the linker to omit WebRTC's Objective-C categories.

Explicitly set `RUSTFLAGS` to `-Clink-arg=-ObjC` for `aarch64-apple-darwin` in the release job. The existing Windows and Linux flag values remain unchanged.

Fixes #45. The same crash signature is reported in [LiveKit #795](https://github.com/livekit/rust-sdks/issues/795).

## Validation

- Reproduced the crash with the published macOS binary.
- Built the fixed macOS release successfully:
  ```sh
  RUSTFLAGS=-Clink-arg=-ObjC cargo build --release --locked --target aarch64-apple-darwin
  ```
- Passed the same native initialization check with the rebuilt binary: WebRTC reached a local mock LiveKit endpoint and handled its HTTP 401 response without aborting.
- Confirmed with `otool` that the rebuilt binary includes the `stringForAbslStringView:` category method.
- Passed `actionlint` and verified the workflow expression's flag values for all three release targets.

Windows and Linux binaries were not built locally.

---
## Summary by cubic
Fixes macOS arm64 release crashes during WebRTC initialization with `+[NSString stringForAbslStringView:]: unrecognized selector`. The release job previously cleared `RUSTFLAGS`, overriding `.cargo/config.toml` and letting the linker drop WebRTC's Objective-C categories. It now sets `-Clink-arg=-ObjC` for Apple targets, and the workflow rejects the binary before packaging if the `stringForAbslStringView:` category is missing. Windows and Linux flag values are unchanged.

<sup>Written for commit a2e5c7a42c6fbe0aa9c3576fb28ceb1f986485b8. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sysprog21/codetrial/pull/46?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


.github/workflows/check.yml
```diff
@@ -558,13 +558,13 @@ jobs:
...
     env:
-      # Cleared for the same reason `plan-mutants` clears it: the workflow-wide
-      # value is `-Clink-arg=-fuse-ld=mold`, and only the Linux jobs that
-      # install mold can honor it. Two of these three runners cannot have mold
-      # at all, and MSVC hands the flag to `link.exe`, which dies with LNK1117.
-      # Release builds happen once per push rather than once per mutant, so the
-      # linker that is present everywhere is the right trade.
-      RUSTFLAGS: ${{ matrix.target == 'x86_64-pc-windows-msvc' && '-C target-feature=+crt-static' || '' }}
+      # Override the workflow's Linux-only mold flag. Even an empty RUSTFLAGS
+      # overrides .cargo/config.toml, so macOS must repeat -ObjC here or the
```
ColtenOuO:
```
      # Override the workflow-wide Linux-only mold flag with the
      # platform-specific flags required by each release target.
      # Even an empty RUSTFLAGS...
```

Nit: This might make the comment more completely describe what it actually does.

.github/workflows/check.yml
```diff
@@ -558,13 +558,13 @@ jobs:
...
+      # linker omits WebRTC's NSString categories and the first room aborts.
+      RUSTFLAGS: >-
```
jserv, line 570: Nothing in this workflow proves the flag reached the linker. A build without `-ObjC` still links cleanly, so the next edit to this expression republishes a macOS binary that aborts at the first room, and the `build` job is skipped on pull requests (it ran as `skipping` here), so the regression surfaces in a user's download rather than in CI. Worth a follow-up next to the Linux `floor` checks that already guard that leg: a macOS-only step before packaging that greps `otool -oV` on the built binary for `stringForAbslStringView:`.

.github/workflows/check.yml
```diff
@@ -558,13 +558,13 @@ jobs:
...
+        ${{ matrix.target == 'x86_64-pc-windows-msvc' && '-C target-feature=+crt-static'
+            || matrix.target == 'aarch64-apple-darwin' && '-Clink-arg=-ObjC'
```
jserv: `.cargo/config.toml` gates this same flag on `cfg(target_vendor = "apple")` while the workflow gates on one exact triple. An `x86_64-apple-darwin` leg added to the matrix would link without `-ObjC` and ship the crash this fixes, with no build failure to say so. Matching the config's shape costs nothing:

```suggestion
            || contains(matrix.target, '-apple-') && '-Clink-arg=-ObjC'
```

.github/workflows/check.yml
```diff
@@ -558,13 +558,13 @@ jobs:
...
     runs-on: ${{ matrix.os }}
     env:
-      # Cleared for the same reason `plan-mutants` clears it: the workflow-wide
-      # value is `-Clink-arg=-fuse-ld=mold`, and only the Linux jobs that
-      # install mold can honor it. Two of these three runners cannot have mold
-      # at all, and MSVC hands the flag to `link.exe`, which dies with LNK1117.
-      # Release builds happen once per push rather than once per mutant, so the
-      # linker that is present everywhere is the right trade.
-      RUSTFLAGS: ${{ matrix.target == 'x86_64-pc-windows-msvc' && '-C target-feature=+crt-static' || '' }}
+      # Override the workflow's Linux-only mold flag. Even an empty RUSTFLAGS
```
jserv: The replaced text carried why the other two legs cannot honor the workflow-wide flag: MSVC hands `-fuse-ld=mold` to `link.exe` and gets LNK1117, neither non-Linux runner can install mold at all, and losing mold is acceptable because releases build once per push rather than once per mutant. That reasoning is not recoverable from the expression, and it is what stops someone from "fixing" the Windows branch later.

.github/workflows/check.yml
abnormal749: Thanks for pointing this out. I added the check to the macOS build before packaging and tested it against both binaries.
The old binary still contains the selector reference, so a plain grep would pass it too. The check matches the method entry in otool output, which catches the old crashing release and lets the fixed one through.

ColtenOuO (approved):
LGTM. Let's wait for CI to be approved and turn green.

jserv:
Thank @abnormal749 for contributing!

PR #49, opened 2026-09-16
Keep selected editor text visible

moon-jam:
## Summary

Selected text in the interview editor was unable to read because the textarea keeps its text transparent while the selection background covers the syntax-highlight layer underneath.

Set a foreground color for selected text so it remains visible while preserving the existing overlay-based syntax highlighting.

## Sample

Before:
image

After:
image

---
## Summary by cubic
Fixes selected text in the interview editor being unreadable by adding a foreground color to `#editor::selection`. The textarea glyphs stay transparent for the syntax-highlight overlay, but the selection background was covering that layer; now selected text uses the ink color and remains visible.

<sup>Written for commit efbb220581dbc734b7d5cee1a34a27350162c6b1. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sysprog21/codetrial/pull/49?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


web/styles.css
```diff
@@ -862,8 +862,13 @@ p {
   display: none;
 }
 
+/* The textarea's normal glyphs are transparent because #editor-highlight
+   paints the syntax-colored text underneath. During selection, however, the
+   textarea's selection background covers that layer, so selected glyphs need
+   their own visible color. */
```
ColtenOuO:
```suggestion
/* Make textarea text visible when its selection background hides the syntax layer. */
```

The first half repeats the nearby comment explaining that the textarea text is transparent and the highlight layer supplies the visible text. Consider keeping only the selection-specific reasoning.

WDYH

ColtenOuO (commented):
Just a minor suggestion, non-blocking for this PR.

web/styles.css
moon-jam: Yeah, I think this is a good point. I've updated the commit. Thanks!

jserv:
Thank @moon-jam for contributing!

PR #50, opened 2026-09-16
Keep one-letter skills without corrupting others

Disesfgewu:
## What

`unique()` dropped any candidate shorter than two characters, meant to drop
stray punctuation left over from a bad split. It also silently dropped
one-letter skills like `C` and `R` before a candidate could choose them.

Removing that floor exposed a second bug: `clean()`'s leading-strip regex
swallowed a bare digit run, so `5G` lost its `5` and survived as the
fabricated skill `G`.

## Why

- `unique()` now keeps a token only if it has a letter. Punctuation-only
  fragments and bare numbers (`24/7`, `100%`, `2015`, `27001`, `3.14`) are
  dropped, whether typed alone or split from a prefix like `ISO`.
  `ISO 9001` and `802.11ac` are kept.
- Bare part numbers such as `6502` and `8051` are dropped too. This is an
  accepted loss: `8051` and `2025` have the same shape, so a filter that goes
  by shape cannot keep one and drop the other. A named part (`MOS 6502`,
  `Z80`) is kept.
- `clean()` now strips only a real list marker: a bullet, a run of dashes or
  asterisks followed by whitespace, or digits followed by `.` or `)` and then
  whitespace or two letters. Digits are matched with `\p{Nd}`, so a
  full-width `１.` counts like `1.`.
- As a result `5G`, `3D`, `4K` and `9.x Java` stay whole, while `1. Must`,
  `1.Must`, `** Must` and `-- Must` lose thei

## Known limits

- The two-letter rule after `.` is a heuristiits
  marker and `3.js experience` loses its `3.`.
- A dash or asterisk glued to the text `--Mu` is,
  since stripping it would eat into real content such as `-Java`.

## Testing

`node --test tests/browser/document-grounding.test.js`: 35 tests pass, with
new coverage for one-letter skills, digit-leded
markers with and without a space, runs of asterisks, version wildcards,
full-width digits, and bare part numbers. `es

---
## Summary by cubic
Fixes grounding extraction so one-letter skills like `C` and `R` survive, and `5G` is no longer corrupted into `G`. The old length floor dropped stray punctuation but also erased one-letter skills; removing it exposed `clean()` treating bare digit runs as list markers.

**Changes**
- `clean()` now strips only real list markers: bullet glyphs, dash/asterisk runs before whitespace, and digits followed by `.`/`)` and then whitespace or two letters. Requirement lines still lose `1.Must` and `** Must`, while `3D` and `9.x Java` stay whole.
- `unique()` now keeps a token only if it has a letter, so pure punctuation and bare numbers are dropped — `24/7`, `100%`, `27001`, `3.14`, and part numbers like `6502` — while named forms like `ISO 9001`, `802.11ac`, and `MOS 6502` survive.

<sup>Written for commit 77f801b88aa3a95252da88c76be72242e0957a55. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sysprog21/codetrial/pull/50?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


web/document-grounding.js
```diff
@@ -115,12 +115,20 @@ function normalizeLines(text) {
...
 }
 
+// Only a real list marker -- a bullet glyph, or digits immediately followed by
+// "." or ")" -- and then whitespace counts as a prefix to strip. A bare
+// leading digit run does not, because that is also how alphanumeric skills
+// and standards spell themselves: "5G", "3D", "802.11" are not "5", "3", and
+// "802" with a stray marker in front.
 function clean(line) {
-  return line.replace(/^[-*•\d.)\s]+/, "").slice(0, textLimit).trim();
+  return line.replace(/^(?:(?:[-*•]|\d+[.)])\s+)+/, "").slice(0, textLimit).trim();
```
jserv: `parseResume` splits `Skills: C, 1. Rust, * Go, • R` on `,;|` and every fragment keeps its leading space, so the `^` anchor never reaches the marker: this now yields `["C", "1. Rust", "* Go", "• R"]` where the old code gave `["C", "Rust", "Go", "R"]`. The leading space also takes up part of the `textLimit` budget, and bullets with no space after them (`•Built a parser`) are no longer stripped either. The new requirement test misses all of this because `normalizeLines` trims whole lines first; please add a case with markers after a comma.

```suggestion
  return line.trimStart().replace(/^(?:(?:[-*]|\d+[.)])\s+|•\s*)+/, "").slice(0, textLimit).trim();
```

web/document-grounding.js
```diff
@@ -115,12 +115,20 @@ function normalizeLines(text) {
...
 }
 
+// A token surviving as nothing but punctuation is what's left of a bad split
+// (a stray delimiter, an orphaned symbol), not a candidate anyone meant to
+// list -- keep only values with at least one letter or digit.
 function unique(values, max) {
-  return [...new Set(values.map(clean).filter((value) => value.length >= 2))].slice(0, max);
+  return [...new Set(values.map(clean).filter((value) => /[\p{L}\p{N}]/u.test(value)))].slice(0, max);
```
jserv: A lone number is now kept as a skill: `Skills: Python, 1, Rust` or `Skills: Python, 1., Rust` gives `["Python", "1", "Rust"]` (or `"1."`), while the old code dropped both. `\p{N}` accepts any single digit, and `clean` only strips `1.` when whitespace follows it. Keeping a token only when it has a letter or at least two digits still passes `C`, `R`, `5G`, `3D` and `802.11`, and drops `1`, `1.`, `/`, `#` and `&`.

```suggestion
  return [...new Set(values.map(clean).filter((value) => /\p{L}|\p{N}.*\p{N}/u.test(value)))].slice(0, max);
```

jserv (changes requested):
Indent via `make indent` and then squash commits without Claude markers.

Disesfgewu:
> Indent via `make indent` and then squash commits without Claude markers.

Sorry about that! I've run make indent, squashed the commits, and pushed the updates. Could you please take another look when you have a moment? Thanks!

jserv:
> I've run make indent, squashed the commits, and pushed the updates. Could you please take another look when you have a moment?

You didn't. Check git manual carefully for `git rebase`.
Don't be overly polite by including "sorry" in replies; it is not particularly useful during collaboration.

Disesfgewu:
Oops. I check that again.

jserv:
Check [Git Squash Commits: How to Combine Commits Into One](https://www.boot.dev/blog/devops/git-squash-commits) by using `git rebase -i` carefully.

ColtenOuO (commented):
It's recommended to make the commit message title relevant to the actual changes, as this helps others quickly understand the updates when reviewing.

Disesfgewu:
> It's recommended to make the commit message title relevant to the actual changes, as this helps others quickly understand the updates when reviewing.

That's very helpful advice. I'll apply this practice to my future commits.

web/document-grounding.js
```diff
@@ -115,12 +115,20 @@ function normalizeLines(text) {
...
-  return line.replace(/^[-*•\d.)\s]+/, "").slice(0, textLimit).trim();
+  return line.trimStart().replace(/^(?:(?:[-*]|\d+[.)])\s+|•\s*)+/, "").slice(0, textLimit).trim();
 }
 
+// A token surviving as nothing but punctuation is what's left of a bad split
+// (a stray delimiter, an orphaned symbol), not a candidate anyone meant to
+// list -- keep only values with at least one letter or digit.
 function unique(values, max) {
-  return [...new Set(values.map(clean).filter((value) => value.length >= 2))].slice(0, max);
+  return [...new Set(values.map(clean).filter((value) => /[\p{L}\p{N}]/u.test(value)))].slice(0, max);
```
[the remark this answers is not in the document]
Disesfgewu: This issue was fixed in previous commit I think.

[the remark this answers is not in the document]
ColtenOuO: The latest commit fixes marker stripping after leading whitespace, but it does not address the bare-numeric-token issue. unique() is unchanged and still accepts 1 and 1. because both match [\p{L}\p{N}].

ColtenOuO (commented):
As a reminder, once the issue pointed out is fixed, we should add a regression test for it to ensure the bug doesn't come back in future updates.

web/document-grounding.js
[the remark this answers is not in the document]
Disesfgewu: Oops. I get it. I would fix it again and check again.

jserv (changes requested):
Rebase latest `main` branch and refine commit messages properly.

web/document-grounding.js
```diff
@@ -120,7 +120,7 @@ function clean(line) {
 }
```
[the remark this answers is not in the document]
Disesfgewu: Oops. It reverted back again.

web/document-grounding.js
```diff
@@ -115,12 +115,20 @@ function normalizeLines(text) {
...
+// and standards spell themselves: "5G", "3D", "802.11" are not "5", "3", and
+// "802" with a stray marker in front.
 function clean(line) {
-  return line.replace(/^[-*•\d.)\s]+/, "").slice(0, textLimit).trim();
+  return line.trimStart().replace(/^(?:(?:[-*]|\d+[.)])\s+|•\s*)+/, "").slice(0, textLimit).trim();
 }
 
+// A token surviving as nothing but punctuation is what's left of a bad split
+// (a stray delimiter, an orphaned symbol), not a candidate anyone meant to
+// list -- keep only values with at least one letter or digit.
```
ColtenOuO: This comment might be outdated. It should be updated based on the current condition check.

tests/browser/document-grounding.test.js
```diff
@@ -24,6 +24,36 @@ test("rejects unsupported, spoofed, empty, oversized, and invalid UTF-8 files",
...
+  // "5G" loses its "5" and survives as the fabricated skill "G".
+  const resume = await parseGroundingFile(txt("Skills: C, 5G, 3D, 4K, 802.11"), "resume");
+  assert.deepEqual(resume.skills, ["C", "5G", "3D", "4K", "802.11"]);
+});
+
+test("a numbered-list marker is still stripped from a requirement line", async () => {
+  const jd = await parseGroundingFile(txt("1. Must know Rust\n2) Should know Go"), "jd");
+  assert.deepEqual(jd.requirements, ["Must know Rust", "Should know Go"]);
+});
+
```
ColtenOuO, line 48:
```suggestion
test("lone numeric fragments are not kept as skills", async () => {
  const resume = await parseGroundingFile(
    txt("Skills: Python, 1, 1., Rust"),
    "resume",
  );
  assert.deepEqual(resume.skills, ["Python", "Rust"]);
});
```

The tests still don't cover the edge case mentioned earlier. Whenever a bug occurs, it highlights a gap in our current test coverage. We should add a test for this specific scenario to prevent regressions in the future.

ColtenOuO:
Also, I don't think it's a good idea to describe the collaboration process in the commit body. It should focus on what was actually changed. 

This issue is still present in the current commit message.

tests/browser/document-grounding.test.js
Disesfgewu: I would add more tests now to edge cases.

web/document-grounding.js
Disesfgewu: I've updated in the latest commit.

web/document-grounding.js
```diff
@@ -115,12 +115,35 @@ function normalizeLines(text) {
...
 }
 
+// Only a real list marker -- a bullet glyph, or digits immediately followed by
+// "." or ")" -- and then whitespace counts as a prefix to strip. A bare
+// leading digit run does not, because that is also how alphanumeric skills
+// spell themselves: "5G" and "3D" are not "5" and "3" with a stray marker in
+// front.
 function clean(line) {
-  return line.replace(/^[-*•\d.)\s]+/, "").slice(0, textLimit).trim();
+  return line.trimStart().replace(/^(?:(?:[-*]|\d+[.)])\s+|•\s*)+/, "").slice(0, textLimit).trim();
```
jserv: Requiring whitespace after the marker breaks list styles that `main` handled. `1.Must know Rust` and `1.) Must know Rust` now reach the interviewer with the marker still attached, and so do `*Should know Go` and `-- Must know Rust`. Letting a numbered marker be followed by a letter keeps `5G` and `3D` intact, because the digits still need a `.` or `)` right after them:

```suggestion
  return line.trimStart().replace(/^(?:(?:-+|\*)\s+|\d+[.)]+(?=\s|\p{L})\s*|•\s*)+/u, "").slice(0, textLimit).trim();
```

`*Should` and hierarchical markers like `1.1 Must` are still not stripped. Handling them without also eating `*nix` or `802.11 experience` is worth a follow-up with its own tests.

web/document-grounding.js
```diff
@@ -115,12 +115,35 @@ function normalizeLines(text) {
...
+// no such ambiguous split reading and survives, same as a plain digit run
+// with no dot ("9001", "27001") or a letter-bearing token ("5G", "4K"). A bare
+// symbol with no digit at all ("/", "#") is a stray delimiter, not a token
+// anyone meant to list, and a lone digit ("1", "-5") still has none of the
+// above shapes' excuse -- both are dropped.
 function unique(values, max) {
-  return [...new Set(values.map(clean).filter((value) => value.length >= 2))].slice(0, max);
+  return [...new Set(values.map(clean).filter((value) => {
+    if (/\p{L}/u.test(value)) return true;
+    if (/^\d+(\.\d+)+$/.test(value)) return false;
```
jserv: `\d` only matches ASCII digits, but the other checks here use `\p{N}`, so `٣.١٤` and `５.２` skip this check, and the last line then keeps them. The comment above also says `3.14` could come from `Python 3.14` being split on whitespace. `parseResume` only splits on `,`, `;` and `|`, so a bare `3.14` can only come from the source text. The PR description says `802.11` passes through, but this rule drops it, and the new test expects that.

```suggestion
    if (/^\p{Nd}+(?:\.\p{Nd}+)+$/u.test(value)) return false;
```

web/document-grounding.js
```diff
@@ -115,12 +115,35 @@ function normalizeLines(text) {
...
+    if (/\p{N}/u.test(value) && /[/%]/.test(value)) return true;
+    return /\p{N}.*\p{N}/u.test(value);
```
jserv: `/\p{N}.*\p{N}/u` accepts any token with two digits anywhere in it, so `-50`, `(3.14)`, `3.14.`, `1-2` and `2020-2024` all become skills. On `main`, `clean()` reduced every one of them to an empty string. Worse, `Skills: 2020, 2021, 2022, 2023, 2024, 2025, 2026, 2027, Rust` now fills all eight skill slots with years and drops `Rust`. If the goal is to keep plain digit runs like `27001`, match exactly that:

```suggestion
    return /^\p{Nd}{2,}$/u.test(value);
```

tests/browser/document-grounding.test.js
```diff
@@ -24,6 +24,107 @@ test("rejects unsupported, spoofed, empty, oversized, and invalid UTF-8 files",
...
+  // "3.14", "5.2", and "802.11" can't be told apart from a GPA or a version
+  // number split off its software name -- there is no letter or symbol left
+  // to say which one it is, so the whole shape is dropped, standard or not.
+  const resume = await parseGroundingFile(txt("Skills: Python, 3.14, 5.2, 802.11, Go"), "resume");
+  assert.deepEqual(resume.skills, ["Python", "Go"]);
+});
+
+test("a bare multi-digit integer is not treated the same as a digit.digit fragment", async () => {
+  // Unlike "3.14", a plain digit run has no dot to make it read as a split
+  // version number or GPA, so it keeps surviving the way it always has.
```
jserv: "keeps surviving the way it always has" is wrong. On `main`, `clean()` stripped every leading digit, so `27001`, `2015` and `754` all came out empty and were dropped. Keeping bare numbers is new behavior in this PR, so the comment and the PR description should say so. They should also explain why a year like `2015` belongs in a skills list at all.

Disesfgewu: The meaning of this is that if the skill start from “GPA 3.87” for example. It will keep surviving. If it only “3.87” would be filtered.

For second 2015 in next test. The reason that it could keep surviving is “ISO” text in the front. but it still has a problem that we could not figure out if ISO 2015 is a license or not by regular expression. 
So in my opinion to deal with these the only way is using LLM or NLP method to do it but it will cost a lot for token and time latency.

web/document-grounding.js
Disesfgewu: I would check this later.
Thank for your suggestion. :)

tests/browser/document-grounding.test.js
jserv: > So in my opinion to deal with these the only way is using LLM or NLP method to do it but it will cost a lot for token and time latency.

Use deterministic approaches wherever possible, so we can enforce restrictions through predefined rules.

ColtenOuO (commented):
I think some of the newly added tests seem unrelated to this PR (e.g., single isolated skill) and might not be necessary.

While adding tests for different cases can be beneficial, I think we should have a clear rationale for why each change is being made. Just to clarify my earlier point about missing tests: I meant that we still need a test specifically reproducing the scenario where the bug occurred, rather than adding unrelated test cases to this PR.

Adding unrelated tests without a clear purpose might just introduce unnecessary review overhead.

cc. @jserv what do you think?

tests/browser/document-grounding.test.js
Disesfgewu: I think use the deterministic method is more convenient and would check it later.
But so in these case, for example “Skills: ISO 2016, 27087, 202610” something like these input, for deterministic approach would pass due to ISO text here, but we still need to pass the text in real ISO like “ISO 27001”. We still have no way to solve but we use specific way only for this case. 
That’s not a great coding way because if do it, we need to cover all possible cases. 🫠

ColtenOuO:
By the way, the PR description might need to be updated to reflect the latest changes, as some parts appear to be outdated.

jserv:
> I think some of the newly added tests seem unrelated to this PR (e.g., single isolated skill) and might not be necessary.
[...]
> Adding unrelated tests without a clear purpose might just introduce unnecessary review overhead.

For small projects like this, it is acceptable to combine diverse changes into a single pull request, provided that each change has a clear purpose and the overall changes are not extensive enough to make the review difficult, allowing reviewers to still walk through them effectively.

tests/browser/document-grounding.test.js
ColtenOuO: ISO does not carry over to the following values here. `parseResume()` splits `Skills: ISO 9001, 27001, 2015` into three independent candidates, so `27001` and `2015` are currently accepted only because bare multi-digit strings pass the filter.

The deterministic rule does not need to verify whether `ISO 2016` is a real standard. It only needs to require enough context in each candidate: keep a token if it contains a Unicode letter. That accepts `ISO 27001`, `IEEE 754`, `802.11ac`, `5G`, `C,` and `R`, while rejecting ambiguous bare values such as `27001`, `2015`, and `754`.

If a document writes `ISO 27001, 27002`, inferring that the prefix applies to both values would be a separate parsing feature. The current parser does not preserve that relationship, so accepting the second number would be guessing.

jserv: > That’s not a great coding way because if do it, we need to cover all possible cases.

So, consider decoupling this feature in a future pull request. For now, only formatting checks are needed.

jserv (changes requested):
Rebase the current branch onto the upstream default branch and rework the series into functionally minimal commits, folding similar ones and enforcing the project's commit message rules.

ColtenOuO:
Squash the commits touching the same files into one and reword the commit message to reflect the updated state.

jserv (changes requested):
Read https://chris.beams.io/git-commit carefully and enforce the rules.

web/document-grounding.js
```diff
@@ -115,12 +115,37 @@ function normalizeLines(text) {
...
+// A letter is what makes a token legible as a named thing: "ISO 27001" and
+// "IEEE 754" keep the org name that scopes their number, "5G" and "3D" carry
+// their own label, and a plain "5" or "27001" or "2015" split off from any
+// of those carries no such scope. parseResume only splits candidates on ","
+// / ";" / "|", so a bare number always comes from the source listing several
+// values after one shared prefix -- "Skills: ISO 27001, 124141, 2015" -- and
+// once split, there is no way left to tell whether "124141" or "2015" is
+// still part of that standard, a separate one, or an unrelated year. Rather
+// than guess, every letterless token is dropped except the one shape that
+// has no such ambiguity: a digit paired with "/" or "%" ("24/7", "100%"),
```
ColtenOuO: Quick question: why should tokens like `24/7` or `100%` (digits paired with `/` or `%`) be considered skills in the first place?

ColtenOuO (commented):
Please remember to resolve the open comments and add regression tests.

web/document-grounding.js
Disesfgewu: Initially, before narrowing this down to just the parsing/filtering problem, I considered that a digit paired with an extra symbol like / or % was still "digit + something extra" and might deserve to pass, the same way a digit paired with a letter would. On reflection, that's not the right criterion — only digit + letter (a real word/label) should pass, since a symbol doesn't name or scope the digit the way a letter does. I removed the digit+symbol exemption; a skill token now needs a letter of its own to survive.

tests/browser/document-grounding.test.js
```diff
@@ -24,6 +24,107 @@ test("rejects unsupported, spoofed, empty, oversized, and invalid UTF-8 files",
...
+  const resume = await parseGroundingFile(txt("Skills: C, Go, Python, R, Rust"), "resume");
+  assert.deepEqual(resume.skills, ["C", "Go", "Python", "R", "Rust"]);
+});
+
+test("digit-led skills are not mistaken for a numbered-list marker", async () => {
+  // clean()'s leading-marker strip is meant for real list prefixes like "1. "
+  // or "2) ", not for a bare digit run: without the "then punctuation" check,
+  // "5G" loses its "5" and survives as the fabricated skill "G".
+  const resume = await parseGroundingFile(txt("Skills: C, 5G, 3D, 4K, 802.11"), "resume");
+  assert.deepEqual(resume.skills, ["C", "5G", "3D", "4K"]);
```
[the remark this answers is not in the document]
Disesfgewu: It had been fixed to drop '802.11' text whether 'IEEE 802.11' or something like this.

web/document-grounding.js
```diff
@@ -115,12 +115,32 @@ function normalizeLines(text) {
...
 }
 
+// Only a real list marker -- a bullet glyph, or digits immediately followed by
+// "." or ")" -- and then whitespace counts as a prefix to strip. A bare
+// leading digit run does not, because that is also how alphanumeric skills
+// spell themselves: "5G" and "3D" are not "5" and "3" with a stray marker in
+// front.
 function clean(line) {
-  return line.replace(/^[-*•\d.)\s]+/, "").slice(0, textLimit).trim();
+  return line.trimStart().replace(/^(?:(?:[-*]|\d+[.)])\s+|•\s*)+/, "").slice(0, textLimit).trim();
```
ColtenOuO: The previously mentioned issue has not been resolved yet. 

Please make sure it is fixed before resolving the comment.

web/document-grounding.js
```diff
@@ -115,12 +115,32 @@ function normalizeLines(text) {
...
 
+// A token with no letter is a number, or nothing but punctuation, wearing a
+// list item's clothes, and neither one is a skill on its own.
+//
+// A letter is what makes a token legible as a named thing: "ISO 27001" and
+// "IEEE 754" keep the org name that scopes their number, "5G" and "3D" carry
+// their own label, and a plain "5" or "27001" or "2015" split off from any
+// of those carries no such scope. parseResume only splits candidates on ","
+// / ";" / "|", so a bare number always comes from the source listing several
+// values after one shared prefix -- "Skills: ISO 27001, 124141, 2015" -- and
```
ColtenOuO: The claim that a bare number always comes from values following a shared prefix is not guaranteed. 

Skills: 2025, Rust produces the same bare numeric candidate without any prefix. The letter-only policy is reasonable, but the rationale should simply say that a letterless candidate lacks enough context to be retained, regardless of how it arose.

ColtenOuO (changes requested):
There are several strange typos in the PR description. Please clean them up.

web/document-grounding.js
Disesfgewu: I had modified that, please check it again.

web/document-grounding.js
Disesfgewu: Oh, I resolved the wrong one. Unresolved it now.

web/document-grounding.js
```diff
@@ -115,12 +115,42 @@ function normalizeLines(text) {
...
+// shapes that a candidate line must not reach the interviewer still wearing
+// its "1." or "1.)". What still has to hold is the reason the whitespace
+// check existed at all -- keeping "5G" and "3D" intact -- and that only
+// needs the digits to be followed by "." or ")": neither of those two ever
+// follows a digit inside an alphanumeric skill, so requiring one before
+// the marker can be dropped is enough on its own, whitespace or not.
 function clean(line) {
-  return line.replace(/^[-*•\d.)\s]+/, "").slice(0, textLimit).trim();
+  return line.trimStart()
+    .replace(/^(?:(?:-+|\*)\s+|\d+[.)]+(?=\s|\p{L})\s*|•\s*)+/u, "")
```
ColtenOuO: The numbered-marker regex still uses ASCII-only `\d`, even though the surrounding filtering is Unicode-aware.  JD requirements such as `１. Must know Rust` retain their markers as well. 

It may be worth tracking in a follow-up PR using `\p{Nd}+` with regression tests for both resumes and JDs.

ColtenOuO (commented):
The commit message is a bit contradictory, please fix it.

By the way, the previously mentioned issue with the PR description has not been resolved.

Once the commit message and PR description are fixed, this will be good to go from my side.

tests/browser/document-grounding.test.js
```diff
@@ -24,6 +24,154 @@ test("rejects unsupported, spoofed, empty, oversized, and invalid UTF-8 files",
...
+  const jd = await parseGroundingFile(txt("1. Must know Rust\n2) Should know Go"), "jd");
+  assert.deepEqual(jd.requirements, ["Must know Rust", "Should know Go"]);
+});
+
+test("a numbered or dashed marker with no space after it is still stripped", async () => {
+  // Requiring whitespace after every marker dropped the marker's punctuation
+  // but not the marker itself for these common copy-paste shapes, so the
+  // digit case relies on "." or ")" following the digits instead: that still
+  // protects "5G" and "3D", which no "." or ")" ever follows.
+  const jd = await parseGroundingFile(txt("1.Must know Rust\n1.) Should know Go\n-- Must know Python"), "jd");
```
jserv, line 50: The dashed fixture carries a space, so the "dashed marker with no space" half of the name is never exercised: `--Must know Python` is left intact, because the dash branch still requires `\s+`. Use that input and make the branch handle it, or narrow the name to the numbered case this actually covers.

web/document-grounding.js
```diff
@@ -115,12 +115,42 @@ function normalizeLines(text) {
...
+// stripped that freely would eat into real content, so those still only
+// count as a marker once whitespace after them confirms it.
+//
+// A numbered marker does not get that same whitespace requirement, because
+// requiring it drops the marker's punctuation but not the marker itself:
+// "1.Must know Rust" and "1.) Must know Rust" are common enough copy-paste
+// shapes that a candidate line must not reach the interviewer still wearing
+// its "1." or "1.)". What still has to hold is the reason the whitespace
+// check existed at all -- keeping "5G" and "3D" intact -- and that only
+// needs the digits to be followed by "." or ")": neither of those two ever
```
jserv: A `.` does follow a digit inside a real token, in a version wildcard: `9.x Java experience required` and `18.x Node experience` are read as a list marker and reach the interviewer as `x Java experience required`. The old character class had the same hole so nothing regressed, but this is the rule the next reader will trust, and requiring two letters after the punctuation closes it while `1.Must` and `1.Go` still strip (the suggestion on the marker branch below carries it).

web/document-grounding.js
jserv: `\*` has no `+` where the dash branch has `-+`, so a run of asterisks stops matching after the first one and the whole branch fails: `** Must know Rust` and `*** Led migration` now reach the interviewer wearing their markers, where the old character class stripped them. Pasted markdown is where these come from.

```suggestion
    .replace(/^(?:(?:-+|\*+)\s+|\d+[.)]+(?=\s|\p{L}{2})\s*|•\s*)+/u, "")
```

web/document-grounding.js
```diff
@@ -115,12 +115,42 @@ function normalizeLines(text) {
...
+// prefix by parseResume's "," / ";" / "|" split -- "Skills: ISO 27001,
+// 124141, 2015". Either way, there is nothing left to tell whether it is
+// still part of a standard, a separate one, or an unrelated year. Rather
+// than guess, every letterless token is dropped, with no exception: "24/7"
+// and "100%" are the same digit-plus-symbol shape as "-50", "1-2", and
+// "2020-2024", and none of them carry a letter to claim a meaning others
+// would have to guess at.
 function unique(values, max) {
-  return [...new Set(values.map(clean).filter((value) => value.length >= 2))].slice(0, max);
+  return [...new Set(values.map(clean).filter((value) => /\p{L}/u.test(value)))].slice(0, max);
```
jserv, line 162: Every letterless token goes, and on a systems resume that is the part numbers: `6502`, `8051`, `68000`, `802.3` and `99.999%` all cleared the old floor and clear nothing here. The rationale above covers years, percentages and split standards but not bare part numbers, which is the one class the audience named in the first test of this change lists without a letter anywhere near them. Worth saying whether that loss is accepted, since the title's bug (a two-character floor hiding `C` and `R`) does not require it.

web/document-grounding.js
Disesfgewu: I've updated the latest PR description here and the commit message.

tests/browser/document-grounding.test.js
Disesfgewu: I would narrow the name to the numbered case this actually covers.

web/document-grounding.js
Disesfgewu: However, tokens like 8051 or 6502 are format-wise indistinguishable from years (e.g., 2015) or arbitrary numbers without deeper contextual analysis.

Since we currently lack the context to tell them apart, treating letterless tokens uniformly is a trade-off we accept for now. I would update the PR rationale to explicitly call out this limitation.

jserv (changes requested):
Refine commit messages per https://chris.beams.io/git-commit

jserv:
Thank @Disesfgewu for contributing!

PR #55, opened 2026-09-17
Hold replay batches until Retry-After has passed

ColtenOuO:
### Summary

After a 429, the replay feed put the refused batch back at the front of the queue and recorded when the server's rate-limit window closes, but nothing read that value. The next flush was still scheduled a fixed second later, and `sendQueuedBatch` posted without checking the window, so a `Retry-After: 30` was answered with a new request every second until the window passed.

### Changes

- Add `scheduleFlush()`, which waits for whichever is later: the usual one second or the end of the rate-limit window. Both fixed one-second `setTimeout` calls now go through it.
- `sendQueuedBatch` no longer posts while the window is open; it reschedules the flush for the window's end.
- Behaviour change: a flush awaited while the window is still open, such as the interview's last one, now returns without posting instead of drawing another 429.
- Add `tests/browser/replay-feed.test.js`, which runs the feed against mocked timers and `fetch`. After a 429 with `Retry-After: 30`, nothing is posted for the next 29 seconds, and at 30 seconds the kept batch goes out together with the event queued behind it.

---
## Summary by cubic
Holds replay batches until the server's Retry-After window has passed. Previously, after a 429 the feed scheduled the next flush a fixed second later and posted without checking the recorded window, so a `Retry-After: 30` drew new requests every second.

**Bug Fixes**
- Adds `scheduleFlush()`, which waits for the later of the usual one second or the end of the rate-limit window; both fixed one-second timers now use it.
- `sendQueuedBatch` no longer posts while the window is open and reschedules the flush for the window's end.
- A flush awaited while the window is still open now returns without posting instead of drawing another 429.
- An interview's final "ended" event is lost if the tab closes before the window passes.
- Adds `tests/browser/replay-feed.test.js`, which mocks timers and `fetch` to verify nothing posts for 29 seconds after a 429 and the kept batch goes out with the queued event at second 30.

<sup>Written for commit 0779fbdf56f2207ebffe900642cf377c2fca246f. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sysprog21/codetrial/pull/55?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


web/replay-feed.js
```diff
@@ -108,6 +114,10 @@ export function flushReplay() {
 
 async function sendQueuedBatch() {
   if (!replayQueue.length || replayClosed) return;
+  if (Date.now() < retryAfter) {
+    scheduleFlush();
+    return;
+  }
```
jserv, line 122: While the window is open, `flushReplay()` now resolves without posting, and the batch waits on a timer that can run as long as `REPLAY_RETRY_MAX_MS`. That contradicts two comments that still describe the old behavior: `endInterview` says its last event is "sent rather than queued", and the doc above `flushChain` says "the interview's last act awaits one". If a candidate closes the tab on the report page before the timer fires, `ended` and `rounds_final` are never sent. Nothing was delivered before this change either (they drew a 429), but please correct both comments here. A follow-up could make the last flush survive the page leaving, for example a `keepalive` post once the window has passed.

tests/browser/replay-feed.test.js
```diff
@@ -0,0 +1,79 @@
...
+
+    for (let second = 1; second < 30; second += 1) {
+      clock.tick(1000);
+      await replayFeed.flushReplay();
+    }
+    assert.deepEqual(posts, [1], "nor does any timer before the window has passed");
+
+    clock.tick(1000);
+    await replayFeed.flushReplay();
+    assert.deepEqual(posts, [1, 2], "then the kept batch goes, with what queued behind it");
```
jserv, line 73: The test calls `flushReplay()` every simulated second and again at the deadline, so it still passes when `scheduleFlush()` never arms a timer. The explicit flush at 30 seconds sends the batch on its own. Let the fake timer do the work and wait for the event loop instead. With this change the test still passes with the fix, and fails both on `main` and when `scheduleFlush()` is a no-op.

```suggestion
    for (let second = 1; second < 30; second += 1) {
      clock.tick(1000);
      await new Promise(setImmediate);
    }
    assert.deepEqual(posts, [1], "nor does any timer before the window has passed");

    clock.tick(1000);
    await new Promise(setImmediate);
    assert.deepEqual(posts, [1, 2], "then the kept batch goes, with what queued behind it");
```

web/replay-feed.js
ColtenOuO: Corrected both comments: the end-of-interview flush is not awaited, and inside a Retry-After window it sends nothing and leaves the queue to the timer. 

The keepalive flush is left for a follow-up.

tests/browser/replay-feed.test.js
ColtenOuO: Applied. Confirmed it passes with the fix and fails on main and with a no-op `scheduleFlush()`.

jserv:

PR #58, opened 2026-09-17
Keep replay posts alive past the page closing

ColtenOuO:
### Summary

Follow-up to #55. The interview's last replay events (`ended`, `rounds_final`) are flushed as the candidate reaches the report page, which is exactly when they are likely to close the tab, and a plain `fetch` is cancelled when the page goes away. Those events were lost and the replay just stopped.

`keepalive: true` alone did not cover the case it was written for: the flag is decided per batch, and the post carrying the lifecycle events is the one most likely to be over the 64 KiB budget, because the ending flush picks up whatever is still queued (an `editor` snapshot of a large buffer, or a backlog left over from a Retry-After window). A batch is now bounded by encoded size, and the queue is compacted when it has to be.

### Changes

- Replay posts use `keepalive: true`, so a batch already on its way survives the tab closing. This applies to every batch rather than only the last one, because an end that lands inside a Retry-After window goes out on the timer, not from the call that asked for it.
- Batches are bounded by encoded size as well as by the server's 32-event limit, and the flush trigger counts queued bytes, so a large `editor` snapshot no longer sits in the queue until the interview ends.
- An event over the budget on its own is still posted, without the flag. The server accepts a body four times that size, and a batch of nothing would leave the queue to a timer that would take nothing again.
- When the lifecycle events are queued behind more than one post's worth of replay, the queued `editor` and `stage` frames that a later queued frame of the same kind restates are dropped, and nothing else. Both are `Restates` kinds, and `replay_review` and `replay_snapshot` already keep only the newest of each, so the only reader that loses anything is one tailing the interview live as it ends. Every transcript line, test run, avatar transition and lifecycle row is kept, in order.

### What this does not cover

- A queue that does not fit one post even after compaction is split into posts rather than compacted further, and the tail can still be lost with the page. Sending that tail without waiting for the head is not safe as things stand: `seq` is allocated inside the insert and every read orders by it, so a small post that wins the race would hide the head's events behind an `ended` row that `responseWindows` stops at. Ordering by anything else reaches the schema, `replay_tail`'s paging and `web/lib.js`.
- A tab closed while a Retry-After window is still open: the batch is waiting on the timer and has not been posted, so there is no request to keep alive.


web/replay-feed.js
```diff
@@ -121,11 +122,19 @@ async function sendQueuedBatch() {
...
+    // which is the moment a candidate is most likely to leave, and a plain
+    // fetch is cancelled with the page. Every batch rather than only the last,
+    // because the end may be the one inside a Retry-After window and so go
+    // out on the timer rather than from the call that asked for it.
     const response = await fetch(`/api/interviews/${encodeURIComponent(state.interviewId)}/events`, {
       method: "POST",
       headers: { "Content-Type": "application/json" },
-      body: JSON.stringify({ events: batch }),
+      body,
+      keepalive: new TextEncoder().encode(body).length <= REPLAY_KEEPALIVE_MAX_BYTES,
```
jserv, line 137: `keepalive` is decided for the whole batch, and `sendQueuedBatch` takes up to `REPLAY_MAX_BATCH` events with no byte bound. The ending `flushReplay` picks up whatever is still queued (an `editor` snapshot of a large buffer from `runTests`, or a backlog left over from a Retry-After window). That makes the post carrying `ended` and `rounds_final` the one most likely to go over 64 KiB and be sent without the flag, which is exactly the case this change is for. Bound the batch by encoded size so the lifecycle events can go out in a small keepalive post. Splitting alone leaves the tail waiting on the head in `flushChain`, so the tail has to be sent without waiting for the head's response.

ColtenOuO: Byte-bounding the batch makes sense, and I will flush on accumulated bytes rather than only on 32 events so an editor snapshot cannot sit in the queue until the end. 

In my option,  sending the tail without waiting for the head looks unsafe as it stands: `seq` is allocated inside the `insert` and every read orders by it, so a small tail that wins the race hides the head's events from responseWindows.

Ordering by anything else touches the `schema`, replay_tail's paging and `web/lib.js`.

ColtenOuO: A trade-off instead: when the final queue does not fit one keepalive post, drop the superseded editor and stage frames still queued, which are Restates kinds a review read already collapses, so the lifecycle events go out in the same ordered post and nothing is left for a tail to wait on. The only reader that loses anything is one tailing the interview live at the moment it ends, and only when the queue overflows a single post.

It doesn't completely solve the problem, but it's a simple workaround

jserv: > It doesn't completely solve the problem, but it's a simple workaround

Make the compaction rule explicit, though: only queued Restate events that are provably superseded by a later event of the same logical state may be discarded. Do not make this a generic "drop events until it fits" policy. Lifecycle and other non-Restate events must remain lossless and ordered. Restates followed by ended/rounds_final, then verify that the redundant Restates are compacted, the final request stays within the keepalive budget and uses keepalive: true, and the lifecycle events remain present and ordered.

ColtenOuO: Done, and kept to the narrow rule. Compaction drops only queued editor and stage frames that a later queued frame of the same kind restates, and only when the lifecycle events are queued behind more than one post's worth of replay; everything else is untouched and in order, and a queue that still does not fit is split into posts rather than compacted further. Batches are now bounded by encoded size as well as by the 32-event limit, and the flush trigger counts bytes, so an editor snapshot no longer waits in the queue for the end.

web/replay-feed.js
```diff
@@ -51,14 +69,22 @@ let replayWindowOpen = false;
 /// stops when the server says it has heard enough.
 export function recordReplay(kind, payload) {
   if (!recordingEnabled || !state.interviewId || replayClosed) return;
-  replayQueue.push({ v: replayVersion, kind, at: Date.now(), payload });
-  if (replayQueue.length >= REPLAY_MAX_BATCH && Date.now() >= retryAfter) {
+  const event = { v: replayVersion, kind, at: Date.now(), payload };
+  replayQueue.push(event);
+  queuedBytes += eventBytes(event);
+  const full =
+    replayQueue.length >= REPLAY_MAX_BATCH || queuedBodyBytes() > REPLAY_KEEPALIVE_MAX_BYTES;
```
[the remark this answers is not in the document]
ColtenOuO: Known, and not introduced here: the same wait exists whenever any post is in flight, including the 32-event flush this branch already had. Both suggested fixes are the ones this PR concluded are unsafe. Dispatching the final events without waiting for the in-flight response loses the ordering that seq carries: it is allocated inside the insert, every read orders by it, and responseWindows stops at the first ended row, so a small post that wins the race hides events that were stored successfully. An unload drain races the same way, and the Fetch keepalive budget is shared across in-flight requests, so the second one can be rejected outright.

Bounding by bytes narrows the window rather than widening it: a large editor snapshot now leaves on its own flush instead of waiting for the end, so less is queued when the interview finishes. The remaining case is in the PR description under what this does not cover.

tests/browser/replay-feed.test.js
```diff
@@ -77,3 +77,208 @@ test("replay-feed waits out Retry-After", async () => {
...
+
+test("replay-feed keeps a batch of exactly the keepalive budget in one post", async () => {
+  // The separators are one fewer than the events. Counted per event instead,
+  // a body of exactly the budget reads as one byte over it and is split, and
+  // the second post is the one that waits on the first and dies with the page.
+  const clock = fakeClock();
+  const feed = await import("../../web/replay-feed.js?boundary");
+  const posts = [];
+  globalThis.fetch = async (_url, init) => {
+    posts.push(new TextEncoder().encode(init.body).length);
```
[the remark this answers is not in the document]
ColtenOuO: Fixed. The boundary test now captures init.keepalive and asserts the exact-budget post carries it, so the case it exists for is the case it checks.

jserv:

PR #59, opened 2026-09-17
Add auto indentation to the current code editor

CX330Blake:
## Summary

Reduce manual formatting during interviews by preserving the current
line indentation and adding one level after opening delimiters and
Python block statements.

Keep delimiter pairs readable by moving matching closing delimiters onto
a separate line when appropriate. Avoid treating comment-only lines as
block openers so comments do not introduce unexpected indentation.

Add regression tests to cover indentation behavior, comments, and C++
preprocessor directives.

## Known Limitations

Check #61


web/editor.js
```diff
@@ -1,5 +1,29 @@
 const INDENT = "    ";
 
+export function indentNewline(value, start, end, language) {
+  const lineStart = start === 0 ? 0 : value.lastIndexOf("\n", start - 1) + 1;
+  const before = value.slice(lineStart, start);
+  const indentation = before.match(/^[ \t]*/)[0];
+  const opener = before.trimEnd().slice(-1);
```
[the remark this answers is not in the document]
CX330Blake: Good catch. This is indeed an edge case, but I think it's acceptable to leave it as-is for now. Supporting trailing Python comments correctly would require more syntax-aware parsing, and I'd prefer to avoid adding that complexity for a relatively uncommon case in this PR.

web/editor.js
```diff
@@ -1,5 +1,29 @@
...
+  const closer = { "{": "}", "[": "]", "(": ")" }[opener];
+  const nested = Boolean(closer) || (language === "python" && opener === ":");
```
jserv: `opener` is the last non-blank character before the caret, even when that character is inside a comment. Pressing Enter after `# Approach:` in Python or `// setup {` in C or JavaScript adds an indent level to a line that does not open a block, and candidates often write comments like `# Steps:` in an interview. Checking for a line that starts with a comment covers the usual case. A trailing comment such as `x = 1  # note:` still gets the extra indent; fixing that means tokenizing the line, which can be a follow-up.

```suggestion
  const comment = /^[ \t]*(#|\/\/)/.test(before);
  const nested = !comment && (Boolean(closer) || (language === "python" && opener === ":"));
```

tests/browser/editor.test.js
```diff
@@ -1,7 +1,78 @@
...
+      { value: expected, start: expected.length, end: expected.length }
+    );
+  }
+});
+
+test("Enter adds a level after opening delimiters and Python colons", () => {
+  for (const [value, language] of [
+    ["    if (ready) {", "javascript"], 
+    ["    items = [", "python"],
+    ["    call(", "java"],
```
jserv: Two of these rows end in a space, and `scripts/git-pre-commit.sh` runs `git diff --cached --check`, which rejects trailing whitespace.

```suggestion
    ["    if (ready) {", "javascript"],
    ["    items = [", "python"],
    ["    call(", "java"],
```

web/editor.js
CX330Blake: Thanks for catching this! I totally missed this case.

web/editor.js
```diff
@@ -1,5 +1,30 @@
 const INDENT = "    ";
 
+export function indentNewline(value, start, end, language) {
+  const lineStart = start === 0 ? 0 : value.lastIndexOf("\n", start - 1) + 1;
+  const before = value.slice(lineStart, start);
+  const indentation = before.match(/^[ \t]*/)[0];
+  const opener = before.trimEnd().slice(-1);
+  const closer = { "{": "}", "[": "]", "(": ")" }[opener];
+  const comment = /^[ \t]*(#|\/\/)/.test(before);
```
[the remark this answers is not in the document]
CX330Blake: Resolved in 7ccb3eb

web/editor.js
```diff
@@ -1,5 +1,30 @@
 const INDENT = "    ";
 
+export function indentNewline(value, start, end, language) {
+  const lineStart = start === 0 ? 0 : value.lastIndexOf("\n", start - 1) + 1;
+  const before = value.slice(lineStart, start);
+  const indentation = before.match(/^[ \t]*/)[0];
+  const opener = before.trimEnd().slice(-1);
```
ColtenOuO, line 18: This is fine to handle as a follow-up, but I'd suggest mentioning it in the PR description for future reference.

jserv (changes requested):
Read https://chris.beams.io/git-commit carefully and enforce the rules.

web/editor.js
```diff
@@ -1,5 +1,30 @@
...
+  const closer = { "{": "}", "[": "]", "(": ")" }[opener];
+  const comment = /^[ \t]*(#|\/\/|\/\*)/.test(before);
```
ColtenOuO: Moving forward, it might be better to have a dedicated helper that detects the programming language first, rather than trying to cover every edge case with a single regex.

Otherwise, if we also need to account for C++, `#define` might trigger a false positive.

tests/browser/editor.test.js
```diff
@@ -1,7 +1,78 @@
...
+  for (const indentation of ["", "  ", "    ", "\t", "\t  "]) {
+    const value = `previous\n${indentation}call();`;
+    const expected = `${value}\n${indentation}`;
+    assert.deepEqual(
+      indentNewline(value, value.length, value.length, "javascript"),
+      { value: expected, start: expected.length, end: expected.length }
+    );
+  }
+});
+
```
ColtenOuO, line 16: I'd suggest adding some tests for the newly added comment handling.

ex.

```
["    # Steps:", "python"]
["    // setup {", "javascript"]
["if ready: # explain", "python"]
["run(); // setup {", "javascript"]
```

ColtenOuO (commented):
Overall, I think this is heading in the right direction, just left a few small suggestions and minor issues to address.

ColtenOuO (commented):
By the way, some of the intermediate review commits can be squashed together.

web/editor.js
```diff
@@ -1,5 +1,41 @@
 const INDENT = "    ";
 
+// TODO: The current implementation cannot handle cases like:
+// /*
+//  * comment
+//  */
```
ColtenOuO, line 6: Could you help open a follow-up issue to track this? I think it's a significant improvement for the project, and it's definitely worth tracking on the issue tracker so everyone can weigh in and discuss.

ColtenOuO (commented):
Would it be worth adding a browser integration test (actually interact with the textarea via Playwright)? Right now, we can only verify whether individual methods work correctly, but we can't validate if the full end-to-end user behavior in the browser matches our expectations.

It might be worth adding because the entire flow actually involves multiple steps. Testing just one part of those steps might not give us comprehensive coverage.

web/editor.js
CX330Blake: Opened a follow-up issue to track syntax-aware indentation and the remaining comment-related edge cases.

Issue: #61

CX330Blake:
> Would it be worth adding a browser integration test (actually interact with the textarea via Playwright)? Right now, we can only verify whether individual methods work correctly, but we can't validate if the full end-to-end user behavior in the browser matches our expectations.
> 
> It might be worth adding because the entire flow actually involves multiple steps. Testing just one part of those steps might not give us comprehensive coverage.

Added Playwright integration tests that interact with the actual editor textarea, press Enter, and verify indentation and caret position across nine cases. They also check syntax highlighting, line numbers, undo/redo, and code retention when switching languages.

web/editor.js
```diff
@@ -1,5 +1,41 @@
...
+function isCommentOnlyLine(line, language) {
+  if (language === "python") {
+    return /^[ \t]*#/.test(line);
+  }
+  return /^[ \t]*(\/\/|\/\*)/.test(line);
```
ColtenOuO, line 11: `isCommentOnlyLine()` treats any C-like line starting with `/*` as comment-only, even when the comment closes and code follows. 

For example, pressing Enter after `/* guard */ if (ready) {` preserves the current indentation instead of adding one level.

Could we either handle text after `*/` or explicitly track this case in #61?

ColtenOuO (commented):
I think we are almost there.

web/editor.js
CX330Blake: Since this is not a common case people are likely to encounter in normal coding, I think we can leave it as a follow-up. I've added this case to #61 for tracking. Thanks for reviewing!

ColtenOuO (approved):
LGTM, thanks for your contribution and patience!

web/editor.js
```diff
@@ -1,5 +1,41 @@
 const INDENT = "    ";
 
+// TODO: The current implementation cannot handle cases like:
```
jserv: The repository carries no other `TODO` in first-party code, and this one records a limitation the pull request description already states, so once this merges it is a marker nothing tracks. Say what `isCommentOnlyLine` actually does instead: it classifies on the first token of a single line and carries no block comment state, which is why a continuation line inside `/* ... */` reads as code.

web/editor.js
jserv: This matches any line whose first token opens a block comment, even when the comment closes and real code follows, so `/* seed */ if (ready) {` is classified as a comment and the next line loses its level. Requiring the comment to stay open on the line keeps the intended cases and drops this one.

```suggestion
  return /^[ \t]*(\/\/|\/\*(?!.*\*\/))/.test(line);
```

web/editor.js
```diff
@@ -1,5 +1,41 @@
...
+    return /^[ \t]*#/.test(line);
+  }
+  return /^[ \t]*(\/\/|\/\*)/.test(line);
+}
+
+export function indentNewline(value, start, end, language) {
+  const lineStart = start === 0 ? 0 : value.lastIndexOf("\n", start - 1) + 1;
+  const before = value.slice(lineStart, start);
+  const indentation = before.match(/^[ \t]*/)[0];
+  const opener = before.trimEnd().slice(-1);
```
jserv: `opener` reads the raw line, so a trailing comment that ends in a delimiter or a colon adds a level the code never opened: `work(); // {` in JavaScript and `value = 1  # note:` in Python both indent the next line. The description covers the opposite direction (a real opener hidden before a trailing comment); this direction costs the candidate a deletion on every such line, so cutting the comment tail before reading the last token is worth doing here rather than in a follow-up.

CX330Blake:
Replaced `isCommentOnlyLine` with `lastCodeCharacter`, which scans the current line while respecting quoted strings and escapes. This fixes Python block openers followed by comments, false indentation from delimiters inside trailing comments, and code following a closed block comment.

Added unit and Playwright coverage for these cases. Multi-line block comment continuation remains unsupported because the scanner carries no state across lines. I'll update the PR description and the description in #61.

ColtenOuO (approved):
Great work!

jserv:
Thank @CX330Blake for contributing!

PR #69, opened 2026-09-20
Tell the interviewer what the timer reads

### LINKED ISSUE #68: AI interviewer incorrectly reports remaining interview time
### Description
I conducted serval interviews using the 30-minute interview duration, with code questions only.

After approximately 15 minutes had elapsed, the AI interviewer sometimes incorrectly reported the remaining interview time. The issue was observed in two different ways: the AI either reported the incorrect remaining time without being asked, or gave the incorrect time when asked.

When this happened, the AI could also urge me to wrap up the interview and stop writing code, even though approximately 15 minutes actually remained.

### Expected behavior
The AI interviewer should report the same remaining interview time as the interview timer shown to the user.

### Actual behavior
After approximately 15 minutes had elapsed, the interview timer showed approximately 15 minutes remaining, but the AI interviewer reported that only 5 minutes remained.

This occurred in two ways:

1. **The AI reported the incorrect time spontaneously.**
    The AI stated that only 5 minutes remained without being asked.
2. **The AI gave the incorrect time when asked.**
    When I asked how much time was left, the AI answered that only 5 minutes remained.

In some cases, the AI then urged me to stop writing code and start explaining my solution verbally.

### Reproduction
The two observed behaviors are:
1. Wait approximately 15 minutes remain → the AI may spontaneously report that only 5 minutes remain.
2. Wait approximately 15 minutes remain → ask the AI how much time is left → the AI may report that only 5 minutes remain.

### Environment
- OS: Windows
- Access Method: Google Chrome via local server (http://127.0.0.1:3000/)
- Git revision: [a388302](https://github.com/sysprog21/codetrial/commit/a388302d940fd0f0764f3b5a55ab7a5f9960cb9c)

ColtenOuO:
### Summary

Fixes #68. 

The live session has no clock. It is told the duration once in its opening instructions and gets a single `[SYSTEM EVENT]` when five minutes remain; between those two points nothing in the session carries the time, and there was no tool that reported it. Every answer about time in between was a guess, and the figure it guessed was the one figure its instructions name, so a candidate fifteen minutes into a thirty-minute interview was told five minutes remained and hurried into wrapping up.

The server's own warning gate was never the problem: it refuses a `time_warning` until the planned duration less five minutes has actually elapsed, so what the candidate heard at minute fifteen was the model's own invention.

### Changes

- One sentence, from one clock. `timer_line` is the only wording any prompt says the time in, and `minutes_left` is the only place it comes from, so a model told these are the only readings it has can recognize every one of them.
- Every stage direction carries it. The data events are stamped at the exit of `apply_data_event`, so an event added later cannot be the one that leaves the interviewer guessing again; the silence nudge and the proactive review are stamped where they are built, and the greeting, `wrap_up` and the reconnect briefing are stamped where they are sent, since none of those pass through `apply_data_event`.
- The stamp does not ask what the prompt already says. These prompts carry the candidate's own text -- the editor snapshot behind a silence nudge, the failure labels behind a test reaction -- so a stamp that skipped a prompt already reading as stamped was one the candidate could take out of the event by typing the sentence, leaving their figure as the only time in it. The reading goes on unconditionally, and the live prompt supplies the rule for the two numbers that can put in one event: the platform's reading is the last sentence, and the same sentence higher up is the candidate's text.
- The five-minute warning no longer repeats the browser's own figure. `control_time_warning` stops reading `remainingSeconds`, so a stale, throttled or forged packet decides when the interviewer is interrupted and never what it is told the time is; the text names the threshold and the stamp supplies the number.
- `read_editor` returns the reading alongside the editor. A stage direction carries the time it was written at, which is already old when the candidate asks how long they have; this is the one thing the interviewer can ask for at any moment.
- Past the deadline the reading is zero. The spoken warning's floor of one minute belongs to the warning: kept here it would promise, for the whole two-minute grace, a minute the candidate's screen does not have.
- The live prompt says the interviewer has no clock of its own, names those two sources as the only ones, and forbids stating or acting on a remaining time that came from neither. The five-minute rule now also says not to warn before the platform's event, and what that costs a candidate.
- Contract bundle 6, live prompt 3. Bundle 5 joins `SCORABLE_CONTRACTS`, since the rubric and the report schema are unchanged and refusing it would blank the scores on every report saved before this.

The reading is wall time from the agent's own start, which is what the page counts down from too: a pause stops the conversation and not the deadline, as `tickTimer` in `web/interview.js` already says at the other end of that decision. That start is taken the moment the candidate joined, before the room and the Gemini session come up, so the setup is not counted against it.

### Not covered

Between events the reading the interviewer holds can be a few minutes old, and nothing pushes a fresh one during a quiet stretch. `read_editor` gives it a current answer whenever it asks, and the prompt tells it to point at the on-screen timer rather than guess. Whether a periodic clock event is worth the turn it costs is better decided after this has been used.


src/agent.rs
```diff
@@ -1470,6 +1470,30 @@ pub fn spoken_minutes_from_remaining_seconds(remaining_seconds: i64) -> i64 {
...
 
+/// What the candidate's countdown reads, in whole minutes.
+///
+/// Wall time, because the page's deadline is wall time too: a pause stops the
+/// conversation and not the clock, as `tickTimer` in `web/interview.js` says
+/// at the other end of the same decision.
+pub fn minutes_left(state: &RuntimeState) -> i64 {
+    let planned = i64::from(state.coding_minutes + state.behavioral_minutes) * 60;
+    let elapsed = i64::try_from(state.started_at.elapsed().as_secs()).unwrap_or(i64::MAX);
+    spoken_minutes_from_remaining_seconds(planned.saturating_sub(elapsed))
```
jserv: `spoken_minutes_from_remaining_seconds` clamps to a floor of 1, which is what the spoken five-minute warning wants and the opposite of what a clock reading wants: once `elapsed` passes `planned` this keeps returning 1, so every stage direction and every `read_editor` during the two-minute `INTERVIEW_DEADLINE_GRACE` window tells the interviewer a minute is still left while the candidate's screen reads `0:00`. That is the one stretch where it has to be closing rather than pacing for another minute.

```suggestion
    let remaining = planned.saturating_sub(elapsed);
    if remaining <= 0 {
        0
    } else {
        spoken_minutes_from_remaining_seconds(remaining)
    }
```

src/agent/events.rs
```diff
@@ -26,13 +26,20 @@ pub fn apply_data_event(
...
         TOPIC_TEST_RESULTS => apply_test_results(state, payload, since_last_test_reaction_seconds),
         TOPIC_CONTROL => apply_control(state, payload),
         TOPIC_INTEGRITY => apply_integrity(state, payload),
         _ => DataEventResult::default(),
-    }
+    };
+    // Stamped here rather than in each arm, because the reading is the same
+    // fact for all of them and an arm added later would otherwise be the one
+    // stage direction that leaves the interviewer guessing again.
+    result.generate_reply = result
```
jserv, line 40: The stamp is unconditional, so the two prompts that already carry a reading now state it twice. The deferred cold brief goes out with `About 15 minutes remain on the candidate's countdown.` near the front and `TIMER: about 15 minutes remain on the candidate's countdown.` at the end, and `control_time_warning` opens with `Exactly 5 minutes remain` taken from the browser's `remainingSeconds` before the stamp restates it from `started_at.elapsed()`, so a late or throttled warning packet hands the model two different numbers inside one event. Either skip a prompt that already ends with the sentence, or let those two arms carry the reading alone.

src/agent/prompts.rs
```diff
@@ -174,8 +174,16 @@ HOW THE SESSION WORKS
   platform (editor snapshots, silence alerts, time warnings). They are NOT spoken
   by the candidate. Never mention them, never read them aloud — just act on them.
 - Editor snapshots show the candidate's code with line numbers like "12| ...".
-- The interview has a visible countdown timer. You will get a [SYSTEM EVENT] when
-  5 minutes remain; verbally warn the candidate at that point.
+- The interview has a visible countdown timer, and you have no clock of your
+  own. Every [SYSTEM EVENT] ends with "TIMER: about N minutes remain", and
```
jserv, line 178: `Every [SYSTEM EVENT] ends with "TIMER: about N minutes remain"` is not what the code emits. The greeting, `wrap_up` and the reconnect `cold_restart` are sent straight to the model rather than through `apply_data_event`, so none of them carries the stamp, and `cold_restart` and `read_editor_text` both word it as `About N minutes remain on the candidate's countdown.` with no `TIMER:` prefix. A model held to `Those are the only times you know` can discard the briefing's reading as coming from neither source, which is the reconnect case this change exists for, so the sentence is worth making identical everywhere it is emitted.

tests/agent.rs
```diff
@@ -754,6 +754,39 @@ fn unpausing_delivers_the_cold_brief_the_pause_deferred() {
...
+/// so. The model has no clock, so it answered with the only figure its
+/// instructions name -- five minutes -- and started hurrying a candidate who
+/// had fifteen.
+#[test]
+fn a_stage_direction_carries_the_countdown_the_model_cannot_see() {
+    let mut state = RuntimeState {
+        coding_minutes: 30,
+        behavioral_minutes: 0,
+        started_at: std::time::Instant::now() - std::time::Duration::from_secs(15 * 60),
+        paused: true,
```
jserv: `Instant::now() - Duration::from_secs(15 * 60)` panics on a machine that has been up for less than fifteen minutes, which an ordinary CI container often is. `past_the_coding_round` in this file exists for that reason and spends the budget instead of winding the clock back; the same assertion holds without the subtraction, with a word changed in the doc comment above.

```suggestion
        coding_minutes: 15,
        behavioral_minutes: 0,
        paused: true,
```

web/lib.js
```diff
@@ -410,21 +410,22 @@ const textEncoder = new TextEncoder();
...
-/// The report keeps the bundle it claims, so its card still says which prompts
-/// wrote it.
+/// and report schema are the active ones. Bundles 4 and 5 differ from 6 only in
+/// the prompts that produced the report, so what their scores mean is
+/// unchanged, and refusing them would blank the scores on every report saved
+/// before bundle 6. The report keeps the bundle it claims, so its card still
+/// says which prompts wrote it.
 export const SCORABLE_CONTRACTS = [
   ACTIVE_CONTRACT,
+  { ...ACTIVE_CONTRACT, bundleVersion: 5, livePromptVersion: 2 },
```
jserv, line 428: Adding bundle 5 to the scorable list also quietly retires the two browser tests that pin the active contract: `report contract migration preserves legacy and rejects unknown provenance` in `tests/browser/lib.test.js` and `report views identify active and legacy scoring contracts` in `tests/browser/render.test.js` both build their `active` fixture as bundle 5 with live prompt 2, and both keep passing against a legacy entry, so nothing exercises bundle 6 through `sanitizeReport` or the report card any more. Bump those fixtures to bundle 6 and move bundle 5 down beside bundle 4 as a legacy case.

src/agent/events.rs
ColtenOuO: Fixed by making the reading come from one place.

src/agent/prompts.rs
ColtenOuO: Right, The greeting, `wrap_up` and the direct `cold_restart` are stamped now, and `read_editor_text` uses `timer_line` too, so every emitted reading is the same sentence.

src/agent.rs
ColtenOuO: Confirmed it and taken as suggested.

tests/agent.rs
ColtenOuO: Taken as suggested, `coding_minutes: 15` with the clock left alone; the doc comment above it changed with it.

web/lib.js
ColtenOuO: Both fixtures are bundle 6 now, with bundle 5 moved down beside bundle 4 as a legacy case, and `lib.test.js` asserts its scores survive rather than only that it renders. Without that the bump would have retired the two tests that were meant to cover it.

jserv:

PR #72, opened 2026-09-20
Add the whiteboard interview mode

### LINKED ISSUE #65: Add a whiteboard interview mode
I would like to propose a new feature: Whiteboard Interview Mode

## Motivation

Many real algorithm interviews are held at a whiteboard: the candidate explains the approach by drawing, traces an example by hand, and ends with handwritten pseudo-code. CodeTrial only offers the editor-and-tests format today, so there is no way to practice explaining an algorithm without a compiler to lean on.

## Proposal

A separate mode, chosen in the lobby next to the duration and the interview loop. A whiteboard interview has no editor, no starter code and no test runs; it uses the same problem bank, taking only the statement and examples.

### Board

Replaces the editor and test panels. Pen, eraser, undo/redo, clear and a few colors, on a fixed-size scrollable canvas so the interviewer always sees the whole board. Mouse input only for the first version. Strokes are kept as vectors, not pixels.

###  How the interviewer sees it

After each stroke settles (about one second of debounce), the board is exported as a JPEG and sent over the data channel to the agent, which forwards it to Gemini Live as a realtime image. The existing camera frame path is off by default and holds a single slot, so the board gets its own path rather than sharing it. A read_board tool, like read_editor, lets the interviewer ask for the latest board.

### Interview flow

Repeat, Examples (drawn), Approach (drawn), Trace (walk an example through the drawing, in place of Code), Edge cases and complexity (in place of Test and Optimize), and a final handwritten pseudo-code step.

### Report

The transcript plus the final board image, graded on clarity, trace correctness, edge cases, complexity and the pseudo-code, still against the problem's optimal and pitfalls. The report request is text-only today and would need to carry an image.

### How a board reaches the interviewer

```mermaid
flowchart LR
    stroke["Stroke ends"] --> debounce["Debounce about 1s"]
    debounce --> jpeg["Export JPEG"]
    jpeg --> channel["Data channel"]
    channel --> agent["Agent"]
    readboard["read_board tool"] --> agent
    agent --> gemini["Gemini Live, realtime image"]
```

ColtenOuO:
### Summary

A whiteboard interview takes the same problem bank and the same six REACTO steps, and swaps the editor and the test runner for a board. Nothing runs, so correctness is what the candidate can defend by tracing an example across their own drawing, and what the session ends with is pseudo-code they write by hand. The lobby offers it beside the duration and the interview loop.

### How the board reaches the interviewer

Strokes are kept as vectors, not pixels. That is what makes undo, redo and the replay possible at all, and it is the one decision everything else follows from.

A second after the drawing settles, the browser exports the board as a JPEG and sends it on a **LiveKit byte stream**rather than on a data topic: a board is tens of kilobytes and `publish_data` carries a single packet. The agent drains the stream on its own task, so the room loop never parks on a read while the candidate's audio queues behind it, andhands the image to Gemini the way a camera frame is handed over, as a realtime image on the live socket. The camera path is untouched and keeps its single slot.

`read_board` cannot answer with the picture, because a tool response is JSON. It records the ask, the room loop re-sends the board, and the text says what an image cannot: how much is on it and how long ago it was drawn. The same re-send runs after a cold restart, so a replacement session is not left knowing a board exists and never having seen one.

### Grading

The final board is attached to the report request as an inline image, ahead of the prompt and on every schema repair.The report prompt follows the surface: it says nothing ran, that correctness is the trace the candidate walked, and what the phase names meant at a board. When no board arrived, it says so rather than sending the reviewer after an attachment that is not there.

The phase ids are deliberately unchanged. `Coding` is the hand trace, `Test` the cases named against the drawing, `Optimizations` the complexity plus the pseudo-code. A second set of ids would have meant a second evidence vocabulary, asecond report schema and a second rubric to calibrate; what changes instead is what the candidate's checklist calls each step and what the reviewer is told they meant.

### Replay

The recording keeps the drawing as the operations that made it, not as images. One board as a JPEG is past `MAX_REPLAY_EVENT_BYTES` on its own and would spend the whole per-interview budget on a handful of frames; the same board as strokes is a few kilobytes. The review page and the recording template rebuild it with `web/whiteboard.js`, the module the candidate drew on, so a stroke refused while drawing is refused coming back off the wire too. Any moment of the interview can be redrawn, rather than the few that could be photographed.

#### Whiteboard

image

Close: #65 

---
## Summary by cubic
Adds a whiteboard interview mode: same problem bank and REACTO steps, but the editor and test runner are swapped for a drawing board. Nothing runs, so correctness is whatever the candidate defends by tracing an example, and the session ends with handwritten pseudo-code. The lobby offers the mode beside the loop and duration.

- Strokes are stored as vectors, not pixels, which is what makes undo, redo, and replay possible. A settled board is exported as a JPEG and travels to the agent on a LiveKit byte stream, then reaches Gemini as a realtime image like a camera frame.
- `read_board` can't return the picture (a tool response is JSON), so it records the ask, the room loop re-sends the board, and it answers with what an image can't: how much is drawn and how long ago.
- The report attaches the final board as an inline image and tells the reviewer nothing ran and what Coding, Test, and Optimizations meant on a board; when no board arrived, it says so rather than pointing at a missing attachment.
- Recordings keep the drawing as the operations that made it rather than as JPEGs, so the replay page and the recording template redraw any moment with `web/whiteboard.js`, the same module the candidate drew on. Phase ids are unchanged.
- Fixes a startup bug where the whiteboard init read a `const` before its initialization, stopping `init()` before the media preflight and making every whiteboard interview unable to start.

Closes #65.

<sup>Written for commit 88b2724289d5c0058ce1b4dc94967fd13840abcc. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sysprog21/codetrial/pull/72?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


ColtenOuO:
Need some time for testing and self-review; will mark this PR as ready for review once it's set.

ColtenOuO:
Perhaps we will need to record the state of the whiteboard at each stage so that we can provide more information to the user at the end of the interview.

ColtenOuO:
Additionally, pseudocode should be graded separately rather than being combined with optimization like it is currently.

alanhc:
On the concern from #65 about whether Gemini can read a drawn board reliably: one option worth considering is Excalidraw as the board layer.

The drawing tools are not the point; the data model is. An Excalidraw scene is a list of typed elements, and boxes, arrows (bound to the boxes they connect) and text keep their actual strings. So next to the JPEG, the agent can get a plain-text summary of the board, e.g. `box "left" -> box "mid"`, and typed pseudo-code arrives as exact text rather than something the model has to read off pixels. The report prompt can quote it too.

I checked whether it fits the vendoring rules, following the `three-vrm.js` precedent (one esbuild bundle, committed, with a reproduce recipe):

- `@excalidraw/excalidraw@0.18.1` + React 19 bundle into a single ES module: 4.9 MB (1.6 MB gzip), plus 145 KB of CSS. That is with `@excalidraw/mermaid-to-excalidraw` aliased to a stub; without it, 8.5 MB.
- Fonts load from `window.EXCALIDRAW_ASSET_PATH`. The Latin ones are ~0.5 MB. The CJK font (Xiaolai) is 13 MB; I left it out, and Excalidraw then falls back to esm.sh for it. The current CSP refuses those loads (230 console violations in my run, drawing and export unaffected), but it needs either vendoring or silencing.
- Served locally with every non-local request aborted, under the policy from `src/web/policy.rs` (`script-src 'self'`, `style-src 'self'`, nothing inline): it mounts in ~200 ms, freehand, shapes and text all work, and `exportToBlob` gives a ~20 KB JPEG. The only inline bit was setting `EXCALIDRAW_ASSET_PATH`, which moves to a same-origin script.
- A 41-point freehand stroke is ~1.5 KB of JSON, so replay would still journal per-element changes from `onChange` rather than whole scenes.

The costs:

- Far bigger than the 280-line `whiteboard.js`, and it brings React into the page.
- The structured benefit exists only when the candidate uses shapes and the text tool. Freehand pseudo-code is still just points. Whether typed text belongs in a whiteboard interview is a product call: less realistic, much easier to grade.
- Text input opens the integrity side: paste and library/file import would have to be disabled.

Only the board layer changes. The byte stream, `read_board`, the report attachment and the replay plumbing stay as they are. Happy to put together the vendored bundle and the scene-to-text summary, either on top of this branch or as a follow-up once it lands, whichever you prefer.

